PR #11, opened 2026-03-28
feat: Ingestion workflow rework and explore UI

harrietkallberg:
Ingestion workflow rework and frontend updates to the exploration layer


max-andreasen (approved):
Looks good!

PR #12, opened 2026-04-01
fix: spring cleaning of repo (tracking and linter)

max-andreasen:
* Removed unecessary tracking, like DS_Store and some .d.ts / .js.map files, which should not be tracked (also added to .gitignore). 
* Several rounds of linting fixes. Nothing too crazy, mostly just one liners and formatting. 
* Biome / lint is now completely clean!


harrietkallberg (commented):
yey clean code!

PR #18, opened 2026-04-15
fix: endpoints and types refactoring

max-andreasen:
* Shared types (e.g. everything from CourseModel) moved to a shared workspace for cleaner imports. 
* Refactored imports to match the new repo structure. 
* Reworked the CourseModel types (now moved to the shared course.ts file). 
* Cleaned backend-nest for uncecessary configs and files. 
* Fix the backend endpoints to use the shared course.ts types (controller level, when sending back data to frontend). 
* This also means that the backend will now return the full and "correct" data, instead of stitching together the scattered data we had before. Will rework some of the endpoints as well. 
* Some final clean-up and making sure everything is working as intended. 

Work yet to be done in Redux / state, and frontend rendering / visualisation. Will be another PR. 
So, the frontend might be a bit broken right now.


harrietkallberg (commented):
se till att docker build anpassas till nytt scheme

harrietkallberg (dismissed):

PR #28, opened 2026-07-18
Feat/profile page

eddbb:
Functionally and visually overhauls the profile page. Adds transcript importation for automatic population of taken courses, grades, and GPA, alongside written and liked reviews.

Changes:

- Added user_courses table in db/schema.ts with migration for courses imported from transcripts
- Added transcript.parser.ts to extract course codes, grades, and credits from transcript text, with unit tests
  - In Ladok: Transcripts and certificates -> Create -> Official transcript of records -> Define: All completed courses, Select language: English, Include: Codes.
- Added /user/transcript upload, /user/transcript-courses fetch, and /user/transcript-courses/:courseCode delete endpoints
- Extended /user/me to return the user's reviews, liked reviews (with like/dislike counts), and transcript courses
- Added public /course/names batch endpoint for resolving course titles
- Rebuilt ProfileView as a dashboard
- Removed the profile picture upload UI (disabled for now)

Database migration needed.


max-andreasen (dismissed):

harrietkallberg:
@greptile-apps review

harrietkallberg:
LGTM! Small note for future reference: `getUserLikedReviews()` does not explicitly filter by `voteType === "like"`. This is fine for now since dislikes are disabled, but we should add that filter if dislikes are enabled later. Approved!

harrietkallberg:
small note, the greptile comments are handled, but we don't have greptile anymore so it didn't retrigger the reviewer :(

backend-nest/src/user/transcript.parser.ts
```diff
@@ -0,0 +1,61 @@
...
+  courseCode: string;
+  grade: string | null;
+  credits: number | null;
+};
+
+// Matches regular course codes (e.g. SF1624) and degree-project codes (e.g. DA231X)
+const COURSE_CODE_RE = /\b([A-Z]{2,3}\d{4}|[A-Z]{2}\d{3}X)\b/;
+// Fx must be checked before F to avoid partial match
+const GRADE_RE = /\b(A|B|C|D|E|Fx|F|VG|G|P|U)\b/;
+const CREDITS_RE = /(\d+[.,]\d+)\s*hp/i;
```
FilipDimitrijevic97, line 11: 🔴 **Blocker** — `CREDITS_RE` requires a decimal (`\d+[.,]\d+`), so credit values written without one — `30 hp`, `6 hp` — parse as `null`. Degree-project rows like `DA231X Degree Project … 30 hp A` are the common case, and those courses then contribute nothing to "Credits earned" or the GPA weighting even though they were passed. Making the decimal part optional (e.g. `\d+(?:[.,]\d+)?`) would cover both formats. 🙂

backend-nest/src/user/transcript.parser.ts
```diff
@@ -0,0 +1,61 @@
...
+    // ("English A") would be mistaken for grades. Fall back to the whole
+    // block for layouts where the grade precedes the credits.
+    let gradeMatch: RegExpExecArray | null = null;
+    if (creditsMatch) {
+      gradeMatch = GRADE_RE.exec(
+        block.slice(creditsMatch.index + creditsMatch[0].length),
+      );
+    }
+    if (!gradeMatch) {
+      gradeMatch = GRADE_RE.exec(block);
```
FilipDimitrijevic97, line 50: 🔴 **Blocker** — nice touch anchoring the grade search after the credits first; that handles the graded case well. One residual gap though: for an *ungraded / pending* course, `creditsMatch` succeeds but there's no grade after it, so we fall through to `GRADE_RE.exec(block)` on the whole block — which can pick up a standalone letter from the course *title*. KTH titles like `Engelska B`, `Matematik D`, or `Fysik C` are common, so `EN1010 Engelska B 6,0 hp` (no grade) would record grade `B` and inflate the GPA. Could we drop the whole-block fallback, or restrict it so it can't match inside the title? 🙏

backend-nest/src/user/user.service.ts
```diff
@@ -195,6 +281,68 @@ export class UserService {
...
+            userId,
+            courseCode: c.courseCode,
+            grade: c.grade,
+            credits: c.credits ?? undefined,
+          })),
+        )
+        .onConflictDoUpdate({
+          target: [schema.userCourses.userId, schema.userCourses.courseCode],
+          set: {
+            grade: sql`excluded.grade`,
```
FilipDimitrijevic97, line 316: 🔴 **Blocker** — one thing to watch with the upsert: `set: { grade: sql\`excluded.grade\`, credits: sql\`excluded.credits\` }` means a re-upload where a course parses *without* a grade (a partial PDF, or the parser edge cases above) would overwrite the previously stored grade/credits with `null` and lose good data. Could we guard against clobbering — e.g. `grade: sql\`COALESCE(excluded.grade, ${schema.userCourses.grade})\`` so we only overwrite when the incoming value is non-null? Same for credits.

frontend/views/ProfileView.tsx
```diff
@@ -1,30 +1,147 @@
...
+  userLikedReviews,
+  transcriptCourses,
+  courseNames,
+  handleTranscriptUpload,
   handleDeleteAccount,
+  handleDeleteCourse,
+  onClickReview,
 }: ProfileViewProps) {
+  const gpa = calculateGPA(transcriptCourses);
+  const totalCredits = transcriptCourses.reduce(
```
FilipDimitrijevic97, line 130: 🟡 **Non-blocking** — small accounting inconsistency: `totalCredits` here sums *every* transcript course, including failed (F/Fx) and ungraded ones, whereas the GPA calc correctly excludes F. So a transcript with an F (7.5 hp) plus a pending course (7.5 hp) would show "15 hp earned" when 0 were actually passed — the two stats end up disagreeing. Might be worth filtering to passed/graded courses so "earned" reflects real credits. Worth a follow-up since it's user-facing.

frontend/views/ProfileView.tsx
```diff
@@ -1,30 +1,147 @@
...
+  );
+
+  const sortedReviews = [...userReviews].sort(
+    (a, b) => (b.likeCount ?? 0) - (a.likeCount ?? 0),
+  );
+  const sortedLikedReviews = [...userLikedReviews].sort(
+    (a, b) => (b.review.likeCount ?? 0) - (a.review.likeCount ?? 0),
+  );
+
+  const formatCredits = (credits: number | null) =>
```
FilipDimitrijevic97, line 142: 🟡 **Non-blocking** — tiny display nit: `formatCredits` always does `toFixed(1)` ("30.0") while the stat tiles use `formatHp` ("30"), so the same value shows two ways on one screen. Reusing `formatHp` here would keep them consistent. Purely cosmetic. 👍

FilipDimitrijevic97 (commented):
Hello!

I ran an agent review. Have an extra look at it, disregard if it is wrong (which it can be):

Thanks for this — really solid work on the transcript parsing and profile page! 🎉 I left a few comments inline. **Three are blockers** (integer credits parsing as null, the ungraded-course grade fallback, and the upsert overwriting good grades with null) since they can produce incorrect GPA/credit data. **Two are non-blocking** — a "Credits earned" accounting inconsistency and a formatting mismatch  🙌

backend-nest/src/user/transcript.parser.ts
FilipDimitrijevic97: Bad example with gymnasium courses, but could be with `Programming in C / Programmering i C`

max-andreasen (changes requested):
Tror på att fixa merge-konflikter och sen merge efter det

PR #59, opened 2026-09-03
refactor(server): group by domain and enforce layering in lint

max-andreasen:
## Why

Server code was organised **by layer** — `api/routers/`, `services/`, `repositories/` — so a change to one domain meant opening three sibling directories, each full of files about domains you weren't touching.

Nobody edits along the layer axis. There is no task shaped like "modify all six repositories but no services". Every real change is domain-shaped: *"add `credits` to course details"* touches `routers/course.ts`, `services/course.ts`, `repositories/course.ts`. Layer-first grouping optimises for a change that doesn't exist, and costs locality on every change that does.

This matters for us specifically because contributors turn over every term and are mostly not senior. The NestJS backend was easy to navigate largely because `src/course/` held everything for one domain in one folder. That property was lost in #58 — not by tRPC, which imposes no layout, but by an undiscussed folder choice. ADR 0002 never mentions it.

## What

Six domains, one folder each:

```
server/course/{router,service,repository}.ts
server/feedback/{router,service,repository}.ts
server/health/{router,service,repository}.ts
server/reviews/{router,service,repository,service.spec}.ts
server/search/{router,service,repository}.ts
server/user/{router,service,repository,service.spec}.ts

server/api/root.ts    # unchanged role: the registry
server/api/trpc.ts    # context, baseProcedure, protectedProcedure
```

`routers/`, `services/` and `repositories/` are gone.

This also makes the server **consistent with the frontend**, which #58 already organised feature-first (`features/<name>/{api,components,hooks}`). Before this PR the repo taught two contradictory conventions — domain-first on the client, layer-first on the server — with no stated reason for either.

## Layering is now enforced, not just documented

`biome.json` gains two `noRestrictedImports` overrides at **error** severity:

- `server/*/router.ts` may not import `**/repository`, or another domain's router
- `server/*/repository.ts` may not import `**/service` or `**/router`

This restores the compile-time boundary NestJS's module graph used to give us. Previously the rule lived only in `CLAUDE.md`, where nothing checked it.

I verified the rule actually fires by injecting both violations:

```
apps/web/server/course/router.ts:1:22 lint/style/noRestrictedImports
  × Routers must not reach into a repository. Call the domain service instead
    (router -> service -> repository).
apps/web/server/user/repository.ts:1:23 lint/style/noRestrictedImports
  × Repositories are the bottom layer: they may import server/db only, never a
    service or router.
Found 2 errors.
```

Both were then reverted.

## No behaviour change

Pure file moves plus import rewrites; git detected 18 of 20 as renames. Services and repositories kept their directory depth, so `../db`, `../errors` and `../ai` resolve unchanged — only routers moved depth (`../trpc` → `../api/trpc`). Cross-domain service calls are preserved: `search/service.ts` still imports `../course/service`.

`CLAUDE.md` and `AGENTS.md` documented the old layout and are updated to state the domain-first rule and point at the enforcing config.

## Verification

- ✅ All **422 local imports across 215 files resolve** (checked with a resolver script — the check that matters for a file move)
- ✅ `biome check apps/web` clean, 0 errors. The 4 remaining `noImgElement` warnings are pre-existing in `Topbar.tsx` / `app-sidebar.tsx`, untouched here
- ✅ lefthook pre-commit biome check passed

## Not in this PR

Two things discussed but deliberately left out, happy to follow up:

- Restoring the 10 spec files #58 deleted (course, feedback, search, departments, ingest have no tests)

https://claude.ai/code/session_0134fagfUVhaUaJXHH31Q9iE


harrietkallberg:
Approve — and thanks for writing up the why rather than just moving files.

The domain-locality argument is the right one. Nobody edits along the layer axis, and src/course/ holding one domain in one folder was the single best thing about the Nest layout. Losing it wasn't a consequence of tRPC, and you're right that ADR 0002 never mentioned the folder choice at all — that was an undiscussed decision inside a large merge, and it shouldn't have been.

What makes this worth more than a rename is the Biome enforcement. A layering rule that lives only in CLAUDE.md is a suggestion; one that fails the build is a boundary. Good call verifying that it actually fires instead of assuming it.

One small ask before I merge: add a line to ADR 0002 stating that the server is grouped by domain and pointing at the lint override. CLAUDE.md and AGENTS.md are the right places for contributors, but the ADR is where someone goes to ask why — and that's exactly the gap this PR is about.

Follow-up I'd like to take you up on: the 10 spec files #58 removed. Course, feedback, search, departments and ingest have no tests right now, and that's the one regression from the rewrite still open. Separate PR is fine.

harrietkallberg (approved):
with an addition to the adr as specified in comment

PR #61, opened 2026-09-03
feat(web): frontend foundation — design tokens, auth dialog, component test harness

harrietkallberg:
Shared groundwork extracted from the landing implementation so the screen tracks can proceed independently. **No user-visible screen here** — every current page renders exactly as before.

Depends on nothing; merges cleanly alongside #60.

## Why this is separate

The landing PR carried four things every other frontend track needs. Holding them behind a screen would block three developers — in particular, **without the test harness nobody else can write a component test at all**.

## What's here

**Design tokens.** The Course Community palette from the design project lands in `globals.css` as an additive `--cc-*` layer, driven by the existing `next-themes` `.dark` class. Nothing in the shadcn palette changed, so current screens are untouched; new work styles against these and gets light/dark for free.

**`AuthReasonDialog`.** One dialog for every reason a visitor is asked to sign in — `log-in`, `sign-up`, `save-course`, `post-review` — each naming the one thing that needs an account and promising the work in progress survives it. It lives in `features/auth` because `save-course` and `post-review` belong to the course and review screens, so those tracks can import it from `@/features/auth` today.

Offers Google, GitHub and email, which is what `server/auth.ts` actually configures. Sign-in returns the visitor to the exact path and query they were on.

**Component test harness.** `vitest` gains three projects — `server` (node), `logic` (node) and `ui` (jsdom) — plus `jsdom` and testing-library. Splitting them means only the component suites pay for a DOM: server code must never see one, and pure logic would otherwise spend ~40s a run booting jsdom for nothing.

**`DialogContent` gains an optional `overlayClassName`.** The repo default (`bg-black/10` + `backdrop-blur`) washes out anything that must stay legible behind a modal. Default behaviour unchanged.

**`biome.json`** — the cross-domain router rule now also matches `@/server/*/router`. `../*/router` alone only caught relative specifiers, so the alias form bypassed the layering guard added in #59. Verified by injecting both forms: the alias was previously undetected, and is now caught.

## Verified on this branch alone

- 28 tests pass (6 files)
- `tsc --noEmit` clean
- `next build` compiles and type-checks
- `biome check` clean on all touched files

https://claude.ai/code/session_017D1FFA2FQVEYvWoEPuYVJk

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -37,7 +37,9 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
-          node-version: 20
+          # jsdom 30 (component tests) needs >=22.22; undici 8 calls
+          # worker_threads.markAsUncloneable, added in Node 22.10.
+          node-version: 24
       - name: Install dependencies
         run: bun install --frozen-lockfile
       - name: Run tests
--- a/apps/web/app/globals.css
+++ b/apps/web/app/globals.css
@@ -45,6 +45,27 @@
   --radius-2xl: calc(var(--radius) * 1.8);
   --radius-3xl: calc(var(--radius) * 2.2);
   --radius-4xl: calc(var(--radius) * 2.6);
+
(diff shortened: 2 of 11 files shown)


Berget1411:
@greptile

max-andreasen (dismissed):

Issue #75, opened 2026-09-03
G: Review domain switch — target columns, votes, and the review contract migration


harrietkallberg:
Move the review domain onto the target columns, then contract the review half of the expand/contract migration. **This work is currently unowned**: `0006` says *"Keep old application writes valid until B switches the review repository"* and `0009` says *"B must drop these triggers"*, but #64's scope is saved/taken/collections and never mentions reviews. `apps/web/server/reviews/**` appears in no issue's "Files you own".

**Land this before #67.** E reads `examination_distribution` and `approach_theory_percent`, and nothing currently writes either one — see "Why this is urgent" below.

## State today

After #63, legacy and target columns coexist on `reviews`. Four pairs are kept in sync by `reviews_legacy_target_sync`; two are not wired at all:

| legacy column | target column | sync today |
|---|---|---|
| `workload` | `workload_score` | trigger |
| `learning_experience` | `learning_score` | trigger |
| `would_recommend` | `happy_took` | trigger |
| `content` | `message` | trigger |
| `examination_methods` | `examination_distribution` | **none** |
| `theoretical_vs_applied` | `approach_theory_percent` | **none** |

`server/reviews/repository.ts` still writes only legacy names — `ReviewWrite` takes `examinationMethods`, `theoreticalVsApplied`, `workload`, `learningExperience`, `wouldRecommend`, `content`. `review_likes` likewise still takes writes while `review_votes` is populated only by trigger.

## Why this is urgent

`reviews` currently holds **0 rows**, so there is no historical data to convert and no lossy backfill to design. That will not stay true. Every review written through the current path lands only in the legacy columns, and the two unwired ones accumulate data with no path forward — so the contract drop costs real data, and the cost grows until this lands.

Doing it while the table is empty removes the hardest part of the problem entirely.

## In scope

- **Switch the write path.** `server/reviews/{router,service,repository}.ts` write the target columns directly. `ReviewWrite` and `apps/web/types/review.ts` take target field names
[...]

PR #72, opened 2026-09-03
feat(db): add target schema compatibility migrations

FilipDimitrijevic97:
Addresses #63
Depends on #71

## Summary
- add ordered target-schema and FK-index migrations without rewriting migration history
- preserve and backfill saved courses, review votes, review fields, and course search data
- keep legacy writers synchronized through temporary compatibility triggers
- settle referential actions, node defaults, timestamptz behavior, real credits, vector(1536), GIN, and HNSW
- capture KOPPS round.ladokUID and add the safe nullable round_code expansion

## Validation
- TypeScript check passes
- 28/28 tests pass
- Drizzle migration integrity check passes
- lint exits successfully; remaining warnings are pre-existing and outside this PR
- diff and historical-migration immutability checks pass

## Remaining before ready for merge
- merge #71 so this stacked PR shows only A2 changes
- apply and verify on a fresh and data-containing isolated Neon child branch
- rebuild course rounds from KOPPS ladokUID values, then add the contract migration for NOT NULL, composite primary key, and serial id removal
- attach the Neon schema diff

No database migration was applied from this worktree.

PR #76, opened 2026-09-03
chore(db): key course_rounds on its serial id and drop round_code

harrietkallberg:
Amends an approved structural decision, so it carries an ADR rather than a silent edit.

## Why

The target keyed `course_rounds` on `(course_code, round_code)`, where `round_code` was to hold KOPPS `round.ladokUID`. `0006` added the column nullable on purpose and deferred the key change until a source-backed rebuild.

**That rebuild is impossible — the KOPPS API is closed.** The 2320 rows in `course_rounds` are all the round data there will be, `round_code` is NULL on every one of them, and `ladokUID` is unobtainable.

Two facts decide the replacement:

**No key derivable from the retained data identifies a round.** Measured against the live database:

| Candidate key | Colliding groups | Excess rows |
| --- | ---: | ---: |
| `(course_code, start_term)` | 184 | 215 |
| `+ language, tutoring_form, tutoring_time_of_day, is_pu, is_vu` | 77 | 104 |
| every column except `id` | 4 | 4 |

Only `ladokUID` ever separated those 77 groups, so a composite key cannot deliver the guarantee it was chosen for — one round per course per offering.

**A synthetic `round_code` would be the id under another name.** `(course_code, id::text)` is unique by construction, not by meaning: a second column, a NOT NULL and a key swap bought for a constraint that constrains nothing. A generated UUID is worse — it discards the identity the rows already have and isn't reproducible. `0006`'s objection to `id::text` was specifically instability across re-ingestion, which cannot occur with the source closed.

Nothing depends on the current key: no foreign key references `course_rounds`, and no query selects `id` or `round_code` — every query in `server/course/repository.ts` filters by `course_code` and reads descriptive fields.

## Changes

- `schema.ts` — `roundCode` removed, along with the partial unique index that only guarded an always-NULL column. `id: serial().primaryKey()` stays
- `server/ingest/` — drops the `roundCode: r.round.ladokUID` mapping and the `ladokUID` field. That field was made *required* in #72 for data that can now never arrive, so leaving it would fail every parse
- `planned-schema-lucid.json` — column removed, `"primaryKey": ["id"]`, note added
- `current-schema-decisions.md` — the approved decision rewritten as superseded rather than deleted
- `planned-database-formats.md` — both references removed; records that `schema_url` is retained because it embeds KTH's own subgroup slug (`.../subgroup/ht-2025-555/...`), present and distinct on 1831 of 2320 rows and the only real round identifier left
- `docs/adr/0004-course-round-identity.md` — new

## Needs a human follow-up

**The Lucid diagram is the visual authority and still shows `(course_code, round_code)`.** It needs the matching edit; I can't make it. Flagged in the ADR's consequences.

## The drop ships here

`0010_abnormal_shocker.sql` drops the column and the partial unique index `0008` added to guard it:

```sql
DROP INDEX "course_rounds_course_code_round_code_unique";
ALTER TABLE "course_rounds" DROP COLUMN "round_code";
```

It was originally deferred to #75 for batching, but unlike the review columns this has **no prerequisite** — NULL on every row, no foreign key, no query, nothing to rewire — so deferring only left the model and the database disagreeing for no benefit. Raised by Greptile; the finding was right. #75 no longer carries it.

## Validation

- `typecheck` exit 0
- 28/28 tests pass
- lint clean; pre-commit hook passes
- Schema diff against `planned-schema-lucid.json`: `course_rounds` now differs only by the pending `round_code` drop. The 2 remaining discrepancies repo-wide are the known `workload_score` / `learning_score` nullability from #72

No database migration was applied from this branch.

https://claude.ai/code/session_015KWRLRZZ3epsR3iHo8YCve

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -85,7 +85,7 @@ _Avoid_: transcript sync, Ladok scrape, upload (that is the file step, not this)
 A named, ordered group of one app user's saved courses. A course may only join a
 collection its owner has also saved.
 _Today_: no table. Closed by `collections` and `collection_courses`.
-_Avoid_: list, folder, group, playlist
+_Avoid_: comparison, list, folder, group, playlist
 
 ### Reviews
 
--- a/apps/web/server/db/drizzle/0010_abnormal_shocker.sql
+++ b/apps/web/server/db/drizzle/0010_abnormal_shocker.sql
@@ -1,0 +1,2 @@
+DROP INDEX "course_rounds_course_code_round_code_unique";--> statement-breakpoint
+ALTER TABLE "course_rounds" DROP COLUMN "round_code";
--- a/apps/web/server/db/drizzle/meta/0010_snapshot.json
+++ b/apps/web/server/db/drizzle/meta/0010_snapshot.json
@@ -1,0 +1,1964 @@
+{
+  "id": "3642ca47-7c55-4c74-acb8-20c1d1dadc8d",
+  "prevId": "a6dc89d3-c077-4d97-9f1c-aa63eedb8983",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.collection_courses": {
+      "name": "collection_courses",
+      "schema": "",
+      "columns": {
+        "collection_id": {
+          "name": "collection_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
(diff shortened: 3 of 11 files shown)


harrietkallberg:
@greptile-apps review

harrietkallberg:
@greptile-apps review

Issue #64, opened 2026-09-03
B: Saved, taken and collections server domains


harrietkallberg:
Build the server side for the three relationships a user has with a course. **Blocked by #62** for the types; you do **not** need to wait for the migration (#63) — mock the repositories and the tests run today.

## In scope

Three domains, each `server/<domain>/{router,service,repository}.ts` plus `service.spec.ts`:

- **`saved`** — save, unsave, list. Saved state is *the existence of a row*, never a flag. Unsaving must leave taken history and reviews untouched
- **`taken`** — add, remove, list, edit self-reported grade/credits/attendance
- **`collections`** — create, rename, delete, reorder; add and remove courses. **A course may only join a collection if the same user has also saved it** — the composite FKs enforce it in the database, but the service must fail cleanly rather than relying on a constraint violation

Register all three in `server/api/root.ts`.

**Also switch the existing favorites code over**: `user.favorites` and `user.toggleFavorite` become the saved domain, and `features/favorites/api/mutations.ts` follows. Retire the old names in this PR — that is the code half of the rename ADR 0003 describes, and the `_Today_` line in `CONTEXT.md` for **Saved course** is deleted here.

## Not in scope

- **No UI.** The Saved / Collections / Taken screens are the frontend track. Rewiring `features/favorites/api/` to the new procedure names is fine; restyling is not
- No migrations (#63)
- No transcript parsing — that is D. You own the `taken` CRUD; D fills it from a Ladok file
- No graph (C)
- Do not touch `server/db/schema.ts` — if a type is wrong, comment on #62

## Files you own

`apps/web/server/saved/**`, `apps/web/server/taken/**`, `apps/web/server/collections/**`, `apps/web/server/api/root.ts`, and `apps/web/features/favorites/api/**` for the rename.

## Spec

- [`CONTEXT.md`](../blob/dev/CONTEXT.md) → "A user's courses". Note the opening line: **saving, taking and reviewing are three independent relationships and none implies another**. A course can be taken without ever being saved
- [`docs/schema_docs/planned-database-formats.md`](../blob/dev/docs/schema_docs/planned-database-formats.md) → "Saved and user domains"

## Conventions that will be enforced in review

`server/<domain>/{router,service,repository}.ts` — grouped by domain, not by layer. Routers stay thin: validate input, pick `baseProcedure` or `protectedProcedure`, call the service. Services hold the logic. Only repositories import `db`.

[...]

Issue #66, opened 2026-09-03
D: Transcript import — Ladok transcript to taken courses


harrietkallberg:
Turn a Ladok transcript into candidate taken courses that the user confirms before anything is written. **Blocked by #62** for types; the parser itself is independent of everything and can start on fixtures immediately.

## In scope

- **Parse** a Ladok transcript into candidate rows: course code, grade, credits, and when it was taken
- **Match** each candidate to `courses.code`. Only catalogue courses can become taken courses — a code with no catalogue entry is reported back, never invented
- **Confirm step.** Nothing is written until the user accepts. Parsing produces a *proposal*
- **Provenance.** Rows created this way set `transcript_imported_at`. A manually added taken course leaves it null — that is the only thing distinguishing the two
- Test fixtures covering a clean transcript, unknown course codes, an unparseable file, and a re-import over existing rows

## Not in scope

- **No UI.** The upload control and the confirm screen are the frontend track. You expose the procedures
- **No taken-course CRUD.** #64 owns `server/taken/`. You produce confirmed rows and hand them over — coordinate on the service signature, do not build a second write path
- **No grade validation.** Imported grades and credits stay **self-reported data** even when they came from a transcript. Do not treat the file as authoritative
- **Do not store the uploaded file** beyond what parsing needs, and do not log its contents. It is a student's academic record
- No `happy_took` — that belongs to a published review, never to the taken-course row
- No migrations (#63). Do not edit `schema.ts` — comment on #62

## Files you own

`apps/web/server/ingest/transcript/**`, and a route handler if the upload needs multipart.

## Spec

- [`CONTEXT.md`](../blob/dev/CONTEXT.md) → **Transcript import**. Note the `_Avoid_` line: it is not a "sync" and not a "scrape", and "upload" names the file step rather than this
- [`docs/schema_docs/planned-database-formats.md`](../blob/dev/docs/schema_docs/planned-database-formats.md) → "Saved and user domains", the `user_taken_courses` paragraph

## A note on the upload path

Multipart does not go through tRPC in this repo — see `app/api/user/profile-picture/route.ts` for the established pattern, and `CLAUDE.md` for the rule. Parsing and confirming can be tRPC procedures; only the file itself needs a route handler.

## Done when
- A fixture transcript produces the right candidates, with unmatched codes reported rather than dropped
- Nothing is written before confirmation
- Confirmed rows carry `transcript_imported_at`; manual rows do not
- Re-importing does not duplicate rows
[...]

Issue #65, opened 2026-09-03
C: Community graph — placement, backbone anchors and bounded viewport query


harrietkallberg:
Build the server side of the community graph: give every user a persistent place in a shared world space, attach them to a few established neighbours, and answer "what is near me" without loading the whole community.

This is the most algorithmic task in the set and it is fully specified. **Blocked by #62** for types; not blocked by the migration — mock the repository and it tests today.

## In scope

`server/graph/{router,service,repository}.ts` + `service.spec.ts`:

- **Placement on join.** Assign and persist a world `x` / `y`, then connect the new node to roughly **3–5 established anchors**. Prefer outer regions for new users
- **Backbone edges.** Store `(node_user_id, anchor_user_id)`. Direction records placement history — newer node to older anchor. Self-edges rejected
- **Bounded neighbourhood query.** Given a user, select a bounded nearby node set up to a product-defined maximum, plus the backbone edges needed for that set. The maximum is a query policy, **not** a database column
- **Node profile defaults** — `color`, `style`, `signal_style` are required, so placement must supply valid values
- **Effective personalization tier** — derived at read time from review activity. May decay one step per complete six months without a qualifying review

## Not in scope

- **No rendering, no canvas, no React.** The landing hero belongs to the frontend track. You return data; the projection into screen pixels is not yours
- **Never write back a projected position.** World units are not browser pixels. Responsive keep-out adjustments happen client-side and must not reach the database
- **Never overwrite `personalization_tier_earned`.** It stores the *highest ever reached*. Decay is derived; if your code writes a lower value, that is a bug
- **Do not globally reposition established users** when the community grows. Local, controlled evolution — not a continuously optimised global layout
- No friendships, no wormholes, no camera or zoom preferences. All explicitly deferred
- No migrations (#63). Do not edit `schema.ts` — comment on #62 instead

## Files you own

`apps/web/server/graph/**`, and `server/api/root.ts` to register.

## Spec

[`docs/landing_docs/personal-community-viewport.md`](../blob/dev/docs/landing_docs/personal-community-viewport.md) — read all of it. "Backbone semantics", "Bounded rendering" and "Persistence boundary" are the requirements. The projection formula is given for context; it runs on the client, not here.

Vocabulary in [`CONTEXT.md`](../blob/dev/CONTEXT.md) → "The community graph". Two words are easy to confuse and reviewers will check:

- **Anchor** — an established node a joining node attaches to
[...]

PR #79, opened 2026-09-03
feat(server): saved, taken and collections domains (#64)

harrietkallberg:
## Before merging

**Nothing blocks this PR — it is first in the wave-1 merge order (#64 → #75 → #65 → #66).** But three things must be true around it:

1. **The migration has not been applied anywhere.** `apps/web/server/db/drizzle/0011_saved_contract.sql` is committed SQL only; no connection to Neon was made from this branch. Apply it to a Neon branch and run Schema Diff before it reaches production.
2. **`0011` is reserved for this PR.** #75 owns `0012`. If #75 somehow lands first, its migration is the one to renumber, not this one.
3. **Order matters.** `0011` drops the `user_favorites` → `user_saved_courses` sync trigger from `0009` and then the table. That is safe only because this same PR moves every application writer onto `user_saved_courses`. Do not apply `0011` to a database still serving the old code.

The three statements are deliberately unguarded — no `IF EXISTS`, no `CASCADE` — so a missing object or an unexpected dependent fails the migration loudly instead of being swallowed.

## What this adds

Three server domains for the three independent relationships an app user has with a course. Each is `server/<domain>/{router,service,repository}.ts` plus a colocated `service.spec.ts`, and all three are registered in `server/api/root.ts`. All fourteen procedures are `protectedProcedure`, and `api/protected.spec.ts` asserts a visitor is rejected on every one of them.

### `server/saved`

`saved.list`, `saved.save`, `saved.unsave`. Saved state is the existence of a row, never a flag, and both writes are idempotent and report the state the caller ended in — `save`/`unsave` rather than a toggle that hid which state you landed in.

The independence rule is tested directly: `saved/service.spec.ts` runs `unsaveCourse` and then asserts that **every** exported function of `taken/repository` and `reviews/repository` was never called. Taken history and reviews survive an unsave.

### `server/taken`

`taken.list`, `taken.add`, `taken.update`, `taken.remove`, over the `recordTakenCourses` contract pinned with #66:

```ts
export function recordTakenCourses(
  userId: string,
  rows: TakenCourseInput[],
  source: TakenCourseWriteSource,
): Promise<{ inserted: number; updated: number }>;
```

Copied exactly as the issue gave it, `TakenCourseInput` and `TakenCourseWriteSource` included. #66 can import it as-is.

Provenance runs through a named repository parameter rather than a bare nullable date, so the rule is legible both at the seam and in the SQL:

```ts
export type TranscriptProvenance =
  | { mode: "set"; importedAt: Date }
  | { mode: "preserve" };
```

`preserve` omits `transcript_imported_at` from the `ON CONFLICT ... DO UPDATE SET` list entirely, which is what stops a manual edit clearing an imported row's provenance. Both directions are covered in `taken/service.spec.ts`.

### `server/collections`

`collections.list`, `create`, `rename`, `delete`, `reorder`, `addCourse`, `removeCourse`.

A course may only join a collection its owner has also saved. The composite FKs enforce it in the database; `addCourseToCollection` checks first via `saved/service.isCourseSaved` (a service → service cross-domain call) and throws `ForbiddenError`, so the caller gets a clean error rather than a constraint violation.

## Concurrency

Greptile's first pass reproduced three write races. All three were real and all three are fixed and verified:

- **`updateTakenCourse` could resurrect a deleted row.** It checked the row existed, then called the upsert; a `remove` landing in between was silently undone. It is now a single `UPDATE ... RETURNING` in the repository, `NotFoundError` when it affects nothing, and it never inserts. The check is gone because the statement *is* the check. `transcript_imported_at` stays out of the SET list, so provenance still survives a manual edit.
- **Two concurrent `addCourse` calls could claim the same position.** `MAX(position)` and the insert were separate statements, and `(collection_id, position)` is not unique, so both committed. Position allocation moved into the repository as `appendCollectionCourse`, which locks the parent collection row `FOR UPDATE` inside one transaction before reading the last position and inserting. Reads also break a position tie on `course_code`, so an already-duplicated position cannot make a listing unstable.
- **Rapid save/unsave clicks could land the wrong way round.** Two fixes: the optimistic cache write now happens before the first `await` and `toggleSaved` reads the cache at call time (so the second click sees where the first is heading), and writes to one course are queued behind whatever is in flight for that course, with only the last one invalidating. Different courses still go in parallel.

## The favorites → saved retirement

- `user.favorites` and `user.toggleFavorite` are deleted outright.
- `user.me.userFavorites` is now `savedCourseCodes`; every reader is updated. `grep` finds no `userFavorites` or `toggleFavorite` left in `apps/web`.
- `features/favorites/api/mutations.ts` exposes `useSetCourseSaved()`, with `setSaved(courseCode, saved)` and `toggleSaved(courseCode)`.
- `schema.ts` loses the `user_favorites` block and its `InsertUserFavorite` / `SelectUserFavorites` types. Nothing else in that file is touched.
- The three `_Today_` lines this PR closes are deleted from `CONTEXT.md`: **Saved course**, **Taken course**, **Collection**. Every other `_Today_` line is left for its own PR.

## Decisions the issue did not settle

- **`collections.reorder` reorders courses *within* a collection**, not the collections themselves. The `collections` table has no position column; `collection_courses.position` is the only ordering the schema carries, and `CONTEXT.md` defines a Collection as "a named, **ordered** group". Reorder takes the desired order and may be a prefix — any member left out keeps its relative order behind the ones named. A course that is not in the collection is rejected.
- **Another user's collection reads as `NotFoundError`, not `ForbiddenError`.** Ownership is scoped in the repository query, so the service never learns whether someone else's collection id exists. A collection is private; its existence is not a stranger's to discover.
- **`recordTakenCourses` dedupes within a batch, last row winning.** PostgreSQL cannot apply `ON CONFLICT` to the same row twice in one statement, so a transcript listing a course twice would otherwise error. An empty batch writes nothing and returns zeroes.
- **The `inserted` / `updated` counts are read before the write.** Two writers racing on the same `(userId, courseCode)` may both report an insert; the rows themselves stay correct. Documented on the function.
- **`isUserFavorite` on `CourseWithUserInfo` is deliberately left alone.** It is a UI prop threaded through the course-card components, which belong to the frontend track — the issue scopes this PR to the server plus `features/favorites/api/**`. It is the last "favorite" identifier in the codebase and should become `isSaved` when whoever owns those components next touches them.
- **Tests mock the repository module (`vi.mock("./repository")`), following `user/` and `reviews/`,** rather than `server/testing/mock-db.ts`. `mock-db.ts` mocks the Drizzle chain, which is a repository-level tool; the seams this issue names are the services, and "mock the repository, not the database" is what the existing specs do. `mock-db.ts` remains unused.

## Gates

- `bun run test:web` — 72 passed (8 files)
- `npx tsc --noEmit` — clean
- `bun run lint` — clean; the 12 warnings are pre-existing `` warnings in `components/layout/app-sidebar.tsx`

Implements #64. No closing keyword on purpose — the product owner closes the issue.


harrietkallberg:
CLoses issue #64

PR #80, opened 2026-09-03
feat(transcript): Ladok transcript import as proposed taken courses (#66)

harrietkallberg:
## Status

Rebased onto `dev` with **#64 and #65 merged** (`5c388ed`, `bfb7536`). The temporary `server/taken/service.ts` placeholder is **deleted** — this branch no longer touches `server/taken/` at all, and every gate has been re-run against #64's real implementation. Clean rebase, no conflicts left.

#64's merged `TakenCourseInput`, `TakenCourseWriteSource` and `recordTakenCourses` match the pinned contract exactly, so there was **no divergence to report**.

Merge order was **#64 → #75 → #65 → #66**; #65 landed ahead of #75 in practice, so **#75 (PR #81) is the only one still ahead of this**. Nothing in this branch depends on it.

---

Implements #66. (The product owner closes the issue, not this PR.)

## What this does

Turns an uploaded Ladok transcript into **candidate** taken courses that the user confirms before anything is written.

```
POST /api/user/transcript   (multipart, session required)
        │  parse + match — stores nothing, writes nothing
        ▼
   TranscriptProposal { candidates[], unmatched[] }
        │  user picks and may edit
        ▼
tRPC transcript.confirm  ──►  recordTakenCourses(userId, rows,
                                { source: "transcript", importedAt })
```

| File | What it is |
| --- | --- |
| `server/ingest/transcript/parse.ts` | **Pure.** Transcript text → candidate rows. |
| `server/ingest/transcript/match.ts` | **Pure.** Candidates → matched against `courses.code`, plus unmatched. |
| `server/ingest/transcript/service.ts` | Builds the proposal; on confirm, calls `recordTakenCourses`. |
| `server/ingest/transcript/pdf-text.ts` | The only impure step: PDF bytes → text, via `unpdf`. |
| `server/ingest/transcript/upload.ts` | Caps the request body as it streams in. |
| `server/ingest/transcript/router.ts` | `transcript.confirm`, `protectedProcedure`. |
| `app/api/user/transcript/route.ts` | The multipart upload; answers with the proposal. |

## The pinned constraints, and where each is honoured

- **`recordTakenCourses` is the only write path.** Nothing here touches `user_taken_courses`, and nothing imports `server/taken/repository.ts`. `service.ts` calls the pinned contract, once, per confirmation.
- **Nothing is written before confirmation.** `buildTranscriptProposal` is read-only, with a test asserting `recordTakenCourses` is never called by it.
- **Unmatched codes are reported, never invented.** `matchCandidates` returns them under `unmatched`. `confirmTranscriptImport` re-checks the catalogue — the confirmed rows arrive from a browser, and nothing stops a client confirming a code that was never proposed — and throws `NotFoundError`, writing nothing, if a code is not in `courses.code`.
- **The file is not stored and not logged.** The bytes live in memory for one request. `TranscriptParseError` messages are fixed strings, and the PDF library's own error is deliberately **dropped rather than chained**, because a PDF parser's message can quote bytes from the document. There is a test asserting the transcript never appears in an error message.
- **Grades and credits stay self-reported.** No grade list, no check of credits against the catalogue — only length and range bounds on the tRPC input.
- **No `happy_took`**, no migration, no `schema.ts` edit, and exactly one `_Today_` line deleted from `CONTEXT.md`: **Transcript import**.

## Parsing, and why it is shaped this way

`unpdf` (new dependency, `apps/web`) reconstructs lines from the PDF's text layer. I compared it against `pdftotext -layout` on the two real KTH transcripts in `transcript_pdfs/`: `pdftotext` scrambled the row/column alignment on the Swedish one, `unpdf` did not. That is why the seam is *text → rows*, and why the fixtures are `.txt`.

Each of these has a test:

- English (`Completed courses`) and Swedish (`Avslutade kurser`)
- Decimal point (`7.5 hp`) and decimal comma (`7,5 hp`)
- A course name wrapped over several lines, with the scope/grade/date landing on a line of its own
- A table continued **across a page break** — including a course name whose two halves sit on either side of the footer. Greptile found both failure modes here in review: the footer being glued into a course name (it carries the student's **personal identity number**), and then, after the first fix, a wrapped row being truncated at the boundary. Page furniture is now stepped over rather than joined or treated as a stop, and **both** properties are pinned by tests so neither fix can undo the other.
- Stopping at `Summation` / `Summering`, so the grading-scale notes (`Excellent (A), Very Good (B)…`) are never read as course rows
- A row whose trailing columns are unreadable: still reported, with nulls, rather than silently dropped
- An unparseable file, an empty table, and a document too large to be a transcript

Verified end to end against both real transcripts (the Swedish and English editions of the same record): **6/6 rows each, every field correct**, including the wrapped course name.

**Those PDFs are not in this PR.** `transcript_pdfs/` and `*.pdf` are both gitignored. The committed fixtures are anonymised — invented name, invented personal identity number, a different course set.

## Security

A security pass over the whole feature found **no exploitable path** for privacy (nothing stores, logs, or leaks a transcript or a personnummer), authorisation (the user id is only ever `ctx.session.user.id`; both endpoints require a session), or integrity (a client cannot confirm a course the catalogue does not have). Two things it did turn up are fixed here:

1. **The parser could be hung by a crafted PDF.** One course code followed by a long run of text made the row pattern backtrack over a growing string on every line — an authenticated request that never returns. Fixed by a cap on the text handed to the parser, a *bounded* lazy quantifier for the name group in the row regex, a cap on continuation lines per row, and truncation of the fallback name.
2. **An oversized upload was buffered before it could be rejected.** A `Content-Length` pre-check is worth less than it looks: the header is absent on a chunked upload and a lie whenever the client wants it to be, and `formData()` buffers the whole body before `file.size` can be read. Replaced with `capRequestBody`, which counts bytes through a `TransformStream` and errors mid-stream, so an oversized body is abandoned rather than held. It has its own spec — under the cap, over the cap, and a `Content-Length` that undercounts.

`bun audit` reports **no advisory against `unpdf@1.8.1`**. (It reports 30 pre-existing findings elsewhere in the tree — `sharp` via `next`, `uuid` — none introduced here.)

**Residual risk, for the record:** `unpdf` is pdf.js, running in-process on attacker-supplied PDFs. The caps here bound *size*, not parser CPU time on a pathologically structured document. Running extraction in a worker with a timeout would be the defence-in-depth fix; it is out of scope for this PR and worth its own issue if you want it.

## Decisions the issue did not settle

1. **Parsing happens in the upload route, not in a tRPC procedure.** The issue says "parsing and confirming can be tRPC procedures; only the file itself needs a route handler". Splitting them would mean holding the transcript's text between two requests — storing a student's academic record, which the issue forbids. So one request uploads *and* parses and returns the proposal; only confirm is tRPC.
2. **`attendance_year` is the year of the result date.** Ladok prints when a result was reported, not which term the course ran. That year is the closest signal the transcript carries, and it is self-reported and user-editable like the rest.
3. **`attendance_periods` is not set by import.** A transcript has no period column; an import must not pretend to know one. The field is left for manual entry, and `ConfirmedTranscriptRow` is `Omit<TakenCourseInput, "attendancePeriods">` to say so at the type level.
4. **Duplicate handling is the write contract's.** `recordTakenCourses` collapses a course code repeated inside one batch (last row winning), so confirming does not dedupe ahead of it — a second filter with the opposite tie-break would be two rules for one behaviour. `matchCandidates` still keeps one row per code in the *proposal*, so the confirm screen never shows a course the user cannot act on twice. Re-import idempotency is the upsert's guarantee; the test here asserts the arguments, mocking the taken service, as the issue instructs.
5. **New dependency: `unpdf`** (~1 MB, no native deps, built for serverless Node). If you would rather not take it, the seam is `pdf-text.ts` alone — everything else is pure and would not change.

## Two open questions for the product owner

1. **Location.** `server/ingest/transcript/` is what issue #66 assigns ("Files you own"), so the router sits one level deeper than the `server/<domain>/router.ts` layout in `CLAUDE.md`, and inside a folder `CLAUDE.md` documents as the KOPPS→Neon pipeline. Code review flagged this: `CONTEXT.md` treats **Transcript import** as a first-class domain, parallel to **Taken course**. I kept the issue's path rather than improvise across the parallel issues' scope, but **`server/transcript/` is arguably the right home** — say the word and it moves.
2. **`catalogueName` is always `titleEng`**, even for a Swedish transcript, because that is the only title `CourseSummary` carries. The response also returns `transcriptName` (the name as printed, in the transcript's own language), so the confirm screen has both. Whoever builds that screen should decide which to show.

## No UI

Per the issue, the upload control and the confirm screen are the frontend track. This PR exposes the endpoints only.

## Gates

`bun run test:web` (144 passed across the suite; 30 of them this feature's), `npx tsc --noEmit`, `bun run lint` — all green. The 12 lint warnings are pre-existing `` warnings in `components/layout/app-sidebar.tsx`.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -74,8 +74,6 @@ _Avoid_: completed course, course history, my courses, enrolled course
 **Transcript import**:
 Reading a Ladok transcript and turning its rows into taken courses after the
 user confirms them. Imported grades and credits remain self-reported.
-_Today_: not implemented. Closed by
-`user_taken_courses.transcript_imported_at`.
 _Avoid_: transcript sync, Ladok scrape, upload (that is the file step, not this)
 
 **Collection**:
--- a/apps/web/app/api/user/transcript/route.ts
+++ b/apps/web/app/api/user/transcript/route.ts
@@ -4,7 +4,7 @@ import { extractTranscriptText } from "@/server/ingest/transcript/pdf-text";
 import { buildTranscriptProposal } from "@/server/ingest/transcript/service";
 import {
   capRequestBody,
-  TranscriptTooLargeError,
+  isTranscriptTooLarge,
 } from "@/server/ingest/transcript/upload";
 
 export const runtime = "nodejs";
@@ -35,7 +35,7 @@ export async function POST(request: Request) {
   try {
     formData = await capRequestBody(request, MAX_BYTES).formData();
   } catch (error) {
-    if (error instanceof TranscriptTooLargeError) {
+    if (isTranscriptTooLarge(error)) {
       return Response.json({ message: TOO_LARGE }, { status: 413 });
     }
     return Response.json(
--- a/apps/web/app/api/user/transcript/route.ts
+++ b/apps/web/app/api/user/transcript/route.ts
@@ -1,0 +1,64 @@
+import { getAuth } from "@/server/auth";
+import { TranscriptParseError } from "@/server/ingest/transcript/parse";
+import { extractTranscriptText } from "@/server/ingest/transcript/pdf-text";
+import { buildTranscriptProposal } from "@/server/ingest/transcript/service";
+
+export const runtime = "nodejs";
+
+const MAX_BYTES = 4 * 1024 * 1024;
+const TOO_LARGE = "Transcript must be less than 4MB";
+
+/**
+ * Turns an uploaded Ladok transcript into a proposal of taken courses.
+ *
+ * Multipart does not go through tRPC in this repo, so the file arrives here —
+ * and because a transcript may not be stored, the same request that receives
+ * the file also parses it and returns the proposal. Confirming the proposal is
+ * `transcript.confirm` over tRPC; this route writes nothing.
+ *
+ * Neither the file nor its text is persisted or logged.
+ */
+export async function POST(request: Request) {
+  const session = await getAuth().api.getSession({ headers: request.headers });
+  if (!session?.user) {
+    return Response.json({ message: "Unauthorized" }, { status: 401 });
+  }
+
+  // Checked before `formData()` so an oversized upload is refused rather than
+  // buffered into memory first. `file.size` below is the check that counts; a
+  // client can send any Content-Length it likes.
+  const declaredLength = Number(request.headers.get("content-length"));
+  if (declaredLength > MAX_BYTES) {
+    return Response.json({ message: TOO_LARGE }, { status: 413 });
+  }
+
+  const formData = await request.formData();
+  const file = formData.get("file");
+  if (!(file instanceof File) || file.size === 0) {
+    return Response.json(
+      { message: "No transcript provided" },
+      { status: 400 },
+    );
+  }
+  if (file.type !== "application/pdf") {
+    return Response.json(
+      { message: "Upload the transcript PDF that Ladok generates" },
+      { status: 400 },
+    );
+  }
+  if (file.size > MAX_BYTES) {
+    return Response.json({ message: TOO_LARGE }, { status: 413 });
+  }
+
+  try {
+    const text = await extractTranscriptText(
+      new Uint8Array(await file.arrayBuffer()),
+    );
+    return Response.json(await buildTranscriptProposal(text));
+  } catch (error) {
+    if (error instanceof TranscriptParseError) {
+      return Response.json({ message: error.message }, { status: 422 });
+    }
+    throw error;
+  }
+}
--- a/apps/web/app/api/user/transcript/route.ts
+++ b/apps/web/app/api/user/transcript/route.ts
@@ -2,6 +2,10 @@ import { getAuth } from "@/server/auth";
 import { TranscriptParseError } from "@/server/ingest/transcript/parse";
 import { extractTranscriptText } from "@/server/ingest/transcript/pdf-text";
 import { buildTranscriptProposal } from "@/server/ingest/transcript/service";
+import {
+  capRequestBody,
+  TranscriptTooLargeError,
+} from "@/server/ingest/transcript/upload";
 
 export const runtime = "nodejs";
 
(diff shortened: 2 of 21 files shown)


harrietkallberg:
Fixed in `cab1ec3` — the P1 was valid, and worse than a cosmetic name: the Ladok page footer carries the student's **personal identity number**, so a row left incomplete by a page break was ending up with a personnummer, the verification URL and the repeated column header inside its course name — and that name goes back to the browser in the proposal.

`ROW.test(open)` closed a row only once it was complete, which an incomplete row never is. Page furniture now closes it too: the Ladok verification line, address block, page counter and repeated header are recognised, and nothing is absorbed again until the next course code. Only the first line of the footer run has to match, since reaching one closes the row.

The trade-off is deliberate and documented at the function: a row split by a page break keeps null columns and is **still reported**, rather than being completed from the wrong lines. Losing a grade is recoverable and the user can fill it in; leaking a personnummer into a course name is not.

Regression test added, using your reproduction directly (`parse.spec.ts`, "never joins page furniture onto an incomplete row at a page break") — it asserts the personnummer, the `ladok.se` URL, the page counter and the header text never appear in any course name, and pins the exact three rows.

Re-verified end to end against the two real KTH transcripts (Swedish and English): still 6/6 rows, every field correct. `bun run test:web` 58 passed, `npx tsc --noEmit` and `bun run lint` clean.

harrietkallberg:
Fixed in `1c77ee9` — valid, and the better shape. Thank you for both rounds; between them they pinned the two properties that were in tension.

Page furniture is now **stepped over** rather than treated as a boundary, so a wrapped course name rejoins across the break while the footer between its halves is still never joined into a name. Skipped lines do not count against the continuation budget, since a page break is not a wrapped line.

That makes the recognition set load-bearing rather than best-effort, so I added the address block (`KTH Royal Institute of Technology` / `Kungliga Tekniska högskolan`, and the postal-code line) alongside the verification lines — every line of the footer run has to be recognised for the leak from your first finding to stay closed.

Both properties are now pinned by tests, so neither fix can silently undo the other:
- `"rejoins a course row that wraps across a page break"` — your fixture, asserting the full name and all three columns.
- `"never joins page furniture onto an incomplete row at a page break"` — asserts no course name may contain the personnummer, the `ladok.se` URL, the page counter or the header text.

Re-verified end to end against the two real KTH transcripts (Swedish and English): still 6/6 rows, every field correct. `bun run test:web` 59 passed, `npx tsc --noEmit` and `bun run lint` clean.

harrietkallberg:
closes issue #66

Issue #77, opened 2026-09-03
H: Move search and ingest onto course_explore, then contract the courses table


harrietkallberg:
Filed out of #75, which identified this as unowned: *\"`courses.name`, `embedding`, `search_vector` and `embedding_hash` are still live and synced by `courses_explore_target_sync`. Dropping them needs `server/search` and `server/ingest` moved onto `course_explore` first — that is a separate unowned issue and should be filed.\"*

This is the **course half** of the expand/contract migration, the mirror of what #75 does for reviews and #64 does for saved. It is mechanical: move the readers and writers, then drop the legacy columns.

## State today

Migration `0006` created `course_explore` as the one-to-one derived-search table. `0009` added trigger `courses_explore_target_sync` + function `sync_courses_to_course_explore`, which copies `courses.embedding`, `courses.embedding_hash` → `source_hash`, and `courses.search_vector` into `course_explore` on every write to `courses`.

Nothing reads `course_explore` yet. `server/search/repository.ts` and `server/ingest/ingest.ts` both still work directly against the legacy columns on `courses`.

## In scope

- **Move the readers.** `server/search/repository.ts` selects `embedding` and `search_vector` from `course_explore`, joined on `course_code`, instead of from `courses`
- **Move the writer.** `server/ingest/` writes embeddings to `course_explore` (`embedding`, `source_hash`, `embedding_model`, `embedded_at`) rather than to `courses.embedding` / `courses.embedding_hash`. `embedding_model` and `embedded_at` exist to make a generated value traceable — populate them, do not leave them null
- **`search_vector` stops being Postgres-maintained.** It is `generatedAlwaysAs` on `courses` today; on `course_explore` the target declares a plain nullable `tsvector`, so the ingest pass must write it. This is noted as a settled decision in #74
- **Drop the trigger** `courses_explore_target_sync` and function `sync_courses_to_course_explore` once nothing writes the legacy columns
- **Contract migration** for the course half: drop `courses.name`, `courses.embedding`, `courses.search_vector`, `courses.embedding_hash`, and the indexes that hang off them. Update `schema.ts` to match
- Vector and GIN indexes move to `course_explore` with the data

Verify the exact legacy column inventory against `docs/schema_docs/planned-schema-lucid.json` before dropping — the target `courses` entry is authoritative for what survives.

## Not in scope

[...]

Issue #68, opened 2026-09-03
F: Frontend — implement the designed pages


harrietkallberg:
Implement the Claude Design artboards. Owned by the one person with design access, so this is the frontend track for the whole group.

**Depends on #61** (design tokens, `AuthReasonDialog`, the jsdom test project). Not blocked by the migration for the pages below.

## Order

1. **App shell** — Page Header and the rail, replacing `AppShell` / `AppSidebar`. First, because every other screen looks wrong until it is right, and it has no data dependency
2. **Course Card** — shared by Explore and Saved
3. **Explore** — the search-and-browse workspace, against the existing `search.courses` / `search.departments`
4. **Review Card**
5. **Saved** — build on today's `user.favorites`; #64 renames it to saved

## Blocked, and on whom

| Page | Waiting for |
|---|---|
| Unreviewed Card | `user_taken_courses` — #62, #64 |
| Taken Courses | #64 |
| Collections | #64 |
| My Page (beyond profile basics) | #64 |
| Find your dot (real positions) | #65 |
| Real card numbers | #67 |

Do not start those; you would be waiting.

## Mock data: use the design's, and shape it to the target

`apps/web/data/courseCardMockData.ts` already backs every chart on the current cards — `getMockChartData`, `getMockCourseStats`, `getMockKeywords`, `getMockSummary`, `getMockPrerequisites`. There is no real aggregation behind any of it (#67).

**Replace its shape with `SAMPLE_COURSE` and `SAMPLE_GEO` from the Course Card artboard.** Those are authoritative: they are the exact fields the design renders, so a card built against them needs no reshaping when #67 lands. Do not invent a third shape.

Then **insulate the swap**: the card takes typed props, and the *screen* maps tRPC output onto them. When #67 arrives you change one mapping function, not the card. This is also what makes cards testable in the `ui` vitest project from #61.

`keywords` and `summary` have **no schema backing at all** and are not in #67. They stay mock until someone decides where they come from — likely generated, since `server/ai.ts` exists. Flag it rather than quietly shipping invented course summaries as if they were real.

## Two conflicts the design surfaced — resolve, do not guess

**1. Score scale.** The design renders workload and learning on **1–5** (`workload: "3.8"`, bar width `76%` = 3.8/5). `docs/schema_docs/current-schema-decisions.md` specifies **1–10 with check constraints**. These contradict. Settle it on #63; the bar maths and axis labels both depend on it.

[...]

Issue #67, opened 2026-09-03
E: Course review aggregation — the numbers the course card renders


harrietkallberg:
The course card renders aggregate review numbers. **None of them exist.** There is no `avg`, `sum` or `count` anywhere in `server/course/repository.ts` or `server/reviews/repository.ts`, and `course.summary` returns catalogue fields only. Every chart and figure on the card today comes from `apps/web/data/courseCardMockData.ts`.

This issue builds the real aggregation. **Blocked by #63** — the source columns change shape in the migration, so building against today's columns means building it twice.

## What the card actually asks for

Taken from `SAMPLE_COURSE` in the design's Course Card artboard, which is the authority on shape:

| Field | Meaning | Source |
|---|---|---|
| `happyPct` | `"87%"` — share of reviewers glad they took it | `reviews.happy_took` |
| `statReviews` | `"148"` — number of reviews | `count(reviews)` |
| `workload` | `"3.8"` — mean workload | `avg(reviews.workload_score)` |
| `learning` | `"4.2"` — mean learning | `avg(reviews.learning_score)` |
| `examLabel` | `"Labs 60% · Exam 40%"` — short examination summary | `reviews.examination_distribution` |
| `statTaken` | `"1.2k"` — how many have taken it | `count(user_taken_courses)` |

Note `examLabel` is a **short summarised string, not the full distribution** — the card shows the top contributors, not a five-way chart. Pick the top two or three and format them.

`statTaken` additionally needs `user_taken_courses`, so it depends on #62 and #64.

## In scope

- Aggregation in `server/course/repository.ts` and `service.ts`, exposed through `course.summary` (and a list variant Explore can call for many courses at once)
- **Batch it.** Explore renders many cards; do not issue one aggregate query per card. A single query keyed by course code returning a map is the shape to aim for
- Handle **no reviews** as a first-class case. The design has explicit `noStats` / `noReviewStats` states: `"No reviews yet"` and `"Be the first to say how it went."` — so return absent, not zero. A course with no reviews must not render as 0% happy
- Exclude nulls correctly: `examination_distribution` and `approach_theory_percent` are nullable because "I don't remember" is an answer. An average must be over reviewers who answered, and the count of answers should be available so the UI can say so
- Tests with mocked repositories covering: no reviews, one review, nulls mixed with answers, and all-null

## Not in scope

- **No UI.** The card is the frontend track. You deliver numbers
[...]

PR #81, opened 2026-09-03
feat(reviews): switch the review domain to the target columns and contract the review migration (#75)

### LINKED ISSUE #75: G: Review domain switch — target columns, votes, and the review contract migration

harrietkallberg:
Closes #75.

## Before merging

**#64 has merged (PR #79) and this branch is rebased on it.** The dependency that section used to describe is discharged:

- `0011_saved_contract` is #64's; `0012_review_contract` is this PR's. The journal's `idx` is now contiguous — the gap this branch was holding open is closed, and `0012`'s `when` is strictly after `0011`'s (drizzle-orm's migrator orders by that timestamp, so an earlier one would have applied `0012` first).
- `meta/0012_snapshot.json` is **regenerated on top of `0011`**. `bun run db:generate` reports "No schema changes, nothing to migrate", and the migration it would otherwise emit is the hand-written `0012` this PR already carries.

**Heads-up for whoever rebases #65 (PR #78) and #66 onto this:** this PR adds a `ValidationError` class to `apps/web/server/errors.ts` and two `instanceof` branches in `apps/web/server/api/trpc.ts`. The change is purely additive, but both files are shared across wave 1, so expect a small conflict there. Raised by #64's agent — see the ownership note further down.

**This migration has not been applied to any database.** No Neon connection was made. `0012_review_contract.sql`, its journal entry and `meta/0012_snapshot.json` are committed SQL only — please apply them to a Neon branch and run Schema Diff before merging.

## What this does

`reviews` holds 0 rows today, so the review half of the expand/contract migration lands with no data to convert. This PR takes that window.

**Write path.** `server/reviews/{router,service,repository}.ts`, `types/review.ts` and the tRPC input all use the target names: `examinationDistribution`, `approachTheoryPercent`, `workloadScore`, `learningScore`, `happyTook`, `message`. Nothing in the app reads or writes a legacy review column or `review_likes` any more.

**The two unwired columns now have a writer.** The review form collects an examination distribution (six whole percentages that must add up to 100) and an approach-theory percentage. Both offer "I don't remember", and that answer **stores `null`, never zeroes** — the invariant `CONTEXT.md` names, defended by tests at the service seam.

**Votes.** Reads and writes move from `review_likes` (`vote_type` text `"like"`/`"dislike"`) to `review_votes` (`voter_user_id`, `up`/`down` enum). Downvoting works again instead of showing "Dislike is temporarily unavailable". The list query tallies `upvoteCount` and `downvoteCount` separately.

A vote is an **idempotent set or remove**, as `planned-database-formats.md` asks. The read only decides which of those the caller meant; the write is a single `INSERT ... ON CONFLICT (voter_user_id, review_id) DO UPDATE` or a delete. Greptile caught the earlier read-then-insert losing a race on a first vote (a double-click was enough to make one request fail with a duplicate-key error after the vote had been saved); a test now races two first votes through a repository double that enforces the primary key the way Postgres does.

**Contract migration `0012_review_contract.sql`.** Drops `reviews_legacy_target_sync` + `sync_legacy_review_fields_to_target` and `review_likes_target_sync` + `sync_review_likes_to_votes`; sets `workload_score` and `learning_score` `NOT NULL`; drops `happy_took`'s temporary `DEFAULT NULL` (0009 added it only so the trigger could tell an omitted target value from an explicit one); drops the six legacy review columns; drops `review_likes`. A `DO $$` pre-flight aborts with a readable message if any row still lacks a target score, rather than letting `SET NOT NULL` fail opaquely. `review_likes` is dropped **without** `CASCADE` so an unexpected dependant aborts the migration instead of being dropped with it.

**Declared target restored.** `schema.ts` declares `workloadScore` / `learningScore` `.notNull()` again, matching `planned-schema-lucid.json`. `meta/0012_snapshot.json` reproduces the `reviews` entry exactly: 11 columns, correct nullability, all three check constraints.

**Scores stay 1–10.** No conversion, no scale parameter. The design's 1–5 is a rendering choice and belongs to #68.

## Please read: things outside this issue's stated file ownership

Two changes were necessary but fall outside `apps/web/server/reviews/**`:

1. **`server/search` — `averageRatings` and the star filter.** `averageRatings` computed `(AVG(examination_methods) + AVG(theoretical_vs_applied) + AVG(workload) + AVG(learning_experience)) / 4`. All four columns are dropped by `0012`, so leaving it would break search at runtime. It now averages the two surviving axes: `(AVG(workload_score) + AVG(learning_score)) / 2`.

   That moved the aggregate from the legacy 1–5 scale to the stored 1–10 one, while the "Minimum Rating" dropdown still sends **stars, 1–5**. Greptile caught the resulting bug (a four-star search returning courses averaging 4/10) and it is fixed: `searchCourses` converts the star threshold up to score points before comparing, so the comparison happens in the scale the columns actually use, and `search/router.ts` bounds `minRating` to 1–5 so the boundary says which scale it speaks. `server/search/service.spec.ts` covers it.

   **What it measures is still a two-axis blend rather than a four-axis one**, and it was already a dubious metric — it averages workload in, and `CONTEXT.md` is explicit that workload is *not* an overall verdict. I preserved its shape rather than redesign it, because aggregation is **#67**'s. #67 should decide what that filter ought to measure.

2. **`server/errors.ts` + `server/api/trpc.ts` — a new `ValidationError`.** The service needs a typed error for "distribution does not add up to 100" / "score out of 1–10" that maps to `BAD_REQUEST`. Both files are shared with the other wave-1 PRs; the change is purely additive (one class, two `instanceof` branches) but **#65 (PR #78) and #66 should expect a small conflict here when they rebase**. Flagged in "Before merging" above too.

## For #67

Your inputs are unchanged from the plan — `examination_distribution`, `approach_theory_percent`, `workload_score`, `learning_score`, `happy_took` keep exactly the names and shapes `planned-schema-lucid.json` specifies. One thing the plan left open that I had to settle:

**`examination_distribution` is a JSON object keyed `{exam, assignments, labs, projects, seminars, other}`, each a whole percentage 0–100, all six summing to exactly 100.** The keys are the broad examination-method labels from `planned-database-formats.md` ("Exam, Assignments, Labs, Projects, Seminars/participation, Other/unspecified"), so a reviewer's remembered mix is comparable to a classified `course_examinations` row. The shape is validated in the service, per that doc's "Validate the JSON shape and total in the service until a database-level JSON constraint is deliberately specified". If you want a different shape, say so before this merges — the table is empty, so it is free to change now and expensive later.

## For #65

`reviews.created_at` and `reviews_user_id_idx` are untouched, as agreed.

## Housekeeping

- The three stale "B" ownership pointers are corrected. `0006` and `0009` now name #75 (and `0009` names #64 for `user_favorites_target_sync` and the unfiled courses/`course_explore` switch for the third trigger). **Comments only — no executable line in an already-applied migration was touched.** The `schema.ts:513` pointer is gone entirely: the temporary `$inferSelect` override it described no longer exists.
- The five review `_Today_` lines are deleted from `CONTEXT.md` (Upvote/Downvote, Happy took, Workload/Learning score, Examination distribution, Approach theory percent). No others.
- Reviews stay readable by visitors: `list` and `byId` are `baseProcedure`; `create`, `update`, `delete` and `vote` are `protectedProcedure` and take the user from `ctx.session`, never from input.

## Hands off to #77

`courses` → `course_explore` is now filed as **#77** ("H: Move search and ingest onto course_explore, then contract the courses table"). `courses.name`, `embedding`, `search_vector` and `embedding_hash` remain live and synced by `courses_explore_target_sync`, which #77 drops.

**#77 inherits the `server/search/**` changes in this PR** — the rewritten `averageRatings` and the star-to-score threshold conversion — so it should rebase on this before moving search onto `course_explore`.

## Tests

`server/reviews/service.spec.ts` drives the seam with the repository mocked, never the database. It covers a review round-tripping through target columns only, `null` `examinationDistribution` / `approachTheoryPercent` surviving as `null` on both create and update, a zero-filled distribution being rejected, a distribution that does not total 100 being rejected, scores outside 1–10 being rejected, a blank message stored as `null`, votes mapping to `up` / `down` (add, flip, and remove-on-repeat), and two concurrent first votes both succeeding. `server/api/protected.spec.ts` gains a visitor check on `reviews.vote`, and `server/search/service.spec.ts` is new (see the star-scale fix above).

Gates: `bun run test:web` (90 passed, this PR's and #64's together), `npx tsc --noEmit`, `bun run lint` all pass. The 12 lint warnings are pre-existing `noImgElement` warnings in `Topbar.tsx` / `app-sidebar.tsx`, untouched by this PR.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -93,36 +93,27 @@ _Avoid_: rating, comment, post, feedback (see **Feedback form**)
 **Upvote** / **Downvote**:
 An app user's judgement that a review was or was not worth reading. A vote is
 about the review, never about the reviewer.
-_Today_: `review_likes.vote_type`, a text column holding `"like"` or
-`"dislike"`. Closed by `review_votes` with an `up`/`down` enum.
 _Avoid_: like, dislike, helpful, helpful score, karma, points, reaction
 
 **Happy took**:
 Whether the reviewer is glad they took the course. A different question from
 whether they would advise a stranger to take it, and it belongs to the review.
-_Today_: `reviews.would_recommend`. Closed by `reviews.happy_took`.
 _Avoid_: would recommend, satisfaction, overall rating
 
 **Workload score** / **Learning score**:
 Two separate axes on a review: how much work the course was, and how much the
 reviewer got out of it. Neither is an overall verdict.
-_Today_: `reviews.workload` and `reviews.learning_experience`, unconstrained
-integers defaulting to 0. Closed by 1-10 scores with check constraints.
 _Avoid_: difficulty, quality, star rating, overall score
 
 **Examination distribution**:
 A reviewer's recollection of how assessment was split across a course. "I don't
 remember" is an answer and stores null, never zeroes. Distinct from **Course
 examination**, which is source data rather than memory.
-_Today_: the single integer `reviews.examination_methods`. Closed by a nullable
-`examination_distribution`.
 _Avoid_: exam breakdown, assessment mix, examination methods
 
 **Approach theory percent**:
 How theoretical rather than applied the reviewer found the course, on the same
 "I don't remember" rule.
-_Today_: the single integer `reviews.theoretical_vs_applied`. Closed by a
-nullable `approach_theory_percent`.
 _Avoid_: theory rating, theoretical vs applied
 
 ### The community graph
--- a/apps/web/features/courses/components/course.tsx
+++ b/apps/web/features/courses/components/course.tsx
@@ -204,13 +204,14 @@ export async function Course({
                 key={review.id}
                 className="w-full max-w-full"
                 courseCode={details.courseCode}
-                wouldRecommend={review.wouldRecommend}
-                content={review.content}
-                examinationMethods={review.examinationMethods}
-                theoreticalVsApplied={review.theoreticalVsApplied}
-                workload={review.workload}
-                learningExperience={review.learningExperience}
-                likeCount={review.likeCount}
+                happyTook={review.happyTook}
+                message={review.message}
+                examinationDistribution={review.examinationDistribution}
+                approachTheoryPercent={review.approachTheoryPercent}
+                workloadScore={review.workloadScore}
+                learningScore={review.learningScore}
+                upvoteCount={review.upvoteCount}
+                downvoteCount={review.downvoteCount}
                 userVote={review.userVote}
                 postId={review.id}
               />
--- a/apps/web/features/reviews/api/mutations.ts
+++ b/apps/web/features/reviews/api/mutations.ts
@@ -20,12 +20,12 @@ export function useCreateReview() {
   );
 }
 
-export function useLikeReview(courseCode: string) {
+export function useVoteOnReview(courseCode: string) {
   const trpc = useTRPC();
   const queryClient = useQueryClient();
   const router = useRouter();
   return useMutation(
-    trpc.reviews.like.mutationOptions({
+    trpc.reviews.vote.mutationOptions({
       onSuccess: () => {
         void queryClient.invalidateQueries({
           queryKey: trpc.reviews.list.queryKey({ courseCode }),
--- a/apps/web/features/reviews/components/post-action-bar.tsx
+++ b/apps/web/features/reviews/components/post-action-bar.tsx
@@ -4,41 +4,45 @@ import { ThumbsDown, ThumbsUp } from "lucide-react";
 import { Button } from "@/components/ui/button";
 import { ButtonGroup } from "@/components/ui/button-group";
 import { cn } from "@/lib/utils";
+import type { ReviewVoteType } from "@/types";
 
 export type PostActionBarProps = {
   postId: string;
-  likeCount: number;
-  dislikeCount: number;
-  userVote: "like" | "dislike" | null;
-  onPostLike: (postId: string) => void;
-  onPostDislike: (postId: string) => void;
+  upvoteCount: number;
+  downvoteCount: number;
+  userVote: ReviewVoteType | null;
+  onVote: (postId: string, voteType: ReviewVoteType) => void;
 };
 
 export default function PostActionBar(props: Readonly<PostActionBarProps>) {
-  const isLiked = props.userVote === "like";
-  const isDisliked = props.userVote === "dislike";
+  const isUpvoted = props.userVote === "up";
+  const isDownvoted = props.userVote === "down";
 
   return (
     <ButtonGroup>
       <Button
-        variant={isLiked ? "secondary" : "ghost"}
-        onClick={() => props.onPostLike(props.postId)}
+        variant={isUpvoted ? "secondary" : "ghost"}
+        aria-pressed={isUpvoted}
+        aria-label="Upvote this review"
+        onClick={() => props.onVote(props.postId, "up")}
       >
         <ThumbsUp
           data-icon="inline-start"
-          className={cn(isLiked && "fill-current")}
+          className={cn(isUpvoted && "fill-current")}
         />
-        {props.likeCount}
+        {props.upvoteCount}
       </Button>
       <Button
-        variant={isDisliked ? "destructive" : "ghost"}
-        onClick={() => props.onPostDislike(props.postId)}
+        variant={isDownvoted ? "destructive" : "ghost"}
+        aria-pressed={isDownvoted}
+        aria-label="Downvote this review"
+        onClick={() => props.onVote(props.postId, "down")}
       >
         <ThumbsDown
           data-icon="inline-start"
-          className={cn(isDisliked && "fill-current")}
+          className={cn(isDownvoted && "fill-current")}
         />
-        {props.dislikeCount}
+        {props.downvoteCount}
       </Button>
     </ButtonGroup>
   );
--- a/apps/web/features/reviews/components/post.tsx
+++ b/apps/web/features/reviews/components/post.tsx
@@ -16,6 +16,7 @@ import type { ExaminationDistribution, ReviewVoteType } from "@/types";
 import {
   EXAMINATION_DISTRIBUTION_KEYS,
   EXAMINATION_DISTRIBUTION_LABELS,
+  MAX_REVIEW_SCORE,
 } from "@/types";
 import { useReviewVotes } from "../hooks/use-review-votes";
 import PostActionBar from "./post-action-bar";
@@ -85,9 +86,6 @@ function truncateHtmlAtWord(html: string, max: number) {
   return result;
 }
 
-/** Scores are stored on a 1-10 scale; see `planned-database-formats.md`. */
-const MAX_SCORE = 10;
-
 function scoreVariant(score: number) {
   if (score >= 8) return "default" as const;
   if (score >= 5) return "outline" as const;
@@ -100,13 +98,13 @@ type ScorePillProps = {
 };
 
 function ScorePill({ name, score }: Readonly<ScorePillProps>) {
-  const aria = `${name}: ${score} out of ${MAX_SCORE}`;
+  const aria = `${name}: ${score} out of ${MAX_REVIEW_SCORE}`;
 
   return (
     <div className="flex items-center gap-2" title={aria}>
       <span className="w-24 text-sm text-muted-foreground">{name}</span>
       <Badge variant={scoreVariant(score)}>
-        {score}/{MAX_SCORE}
+        {score}/{MAX_REVIEW_SCORE}
       </Badge>
     </div>
   );
@@ -156,22 +154,17 @@ type ExaminationSummaryProps = {
 function ExaminationSummary({
   distribution,
 }: Readonly<ExaminationSummaryProps>) {
-  const shares =
-    distribution === null
-      ? []
-      : EXAMINATION_DISTRIBUTION_KEYS.filter(
-          (key) => distribution[key] > 0,
-        ).map((key) => ({ key, share: distribution[key] }));
-
   return (
     <div className="flex flex-wrap items-center gap-2">
       <span className="w-24 text-sm text-muted-foreground">Examination</span>
-      {shares.length === 0 ? (
+      {distribution === null ? (
         <Badge variant="secondary">Not remembered</Badge>
       ) : (
-        shares.map(({ key, share }) => (
+        EXAMINATION_DISTRIBUTION_KEYS.filter(
+          (key) => distribution[key] > 0,
+        ).map((key) => (
           <Badge key={key} variant="outline">
-            {EXAMINATION_DISTRIBUTION_LABELS[key]} {share}%
+            {EXAMINATION_DISTRIBUTION_LABELS[key]} {distribution[key]}%
           </Badge>
         ))
       )}
--- a/apps/web/features/reviews/components/post.tsx
+++ b/apps/web/features/reviews/components/post.tsx
@@ -12,6 +12,11 @@ import {
 import { Separator } from "@/components/ui/separator";
 import { sanitizeHtml } from "@/lib/sanitize-html";
 import { cn } from "@/lib/utils";
+import type { ExaminationDistribution, ReviewVoteType } from "@/types";
+import {
+  EXAMINATION_DISTRIBUTION_KEYS,
+  EXAMINATION_DISTRIBUTION_LABELS,
+} from "@/types";
 import { useReviewVotes } from "../hooks/use-review-votes";
 import PostActionBar from "./post-action-bar";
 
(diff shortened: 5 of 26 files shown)


harrietkallberg:
Thanks — the P1 is real and is fixed in 9a23e06.

Dropping the four legacy 1–5 review columns moved `averageRatings` onto the stored 1–10 scale, while the "Minimum Rating" dropdown still sends stars (1–5), so a four-star search was matching courses averaging 4/10.

I took the first of your two options: convert the threshold rather than the data. `searchCourses` now multiplies the star threshold up to score points before comparing, so the comparison happens in the scale the columns actually use and no rounding is invented on the way down. Per your note about validating the range, `search/router.ts` now bounds `minRating` to `z.number().int().min(1).max(5)`, so the boundary states which scale it speaks, and `searchCourses` documents that its `minRating` is in stars.

I did not standardise the UI on 1–10: the 1–5 star rendering is a deliberate product decision that issue #75 explicitly leaves to #68, and #67 owns what this blended rating should measure at all.

New `apps/web/server/search/service.spec.ts` covers it — a four-star filter excluding a 4/10 course, a three-star filter keeping a 6/10 one and dropping a 5/10 one, and the unfiltered path not calling `averageRatings`. It fails on the previous commit and passes on this one. Gates: 45 tests, `tsc --noEmit`, and `biome check` all pass.

harrietkallberg:
The P1 is real and is fixed in 5c685e9.

`toggleVote` read for an existing vote and then inserted. Two requests racing on a first vote — a double-click is enough — both saw nothing, both inserted, and the composite primary key rejected the second. The vote was saved; that user was told "Failed to update vote."

I took the conflict-aware option. The read now only decides which operation the caller meant; the write is a single `INSERT ... ON CONFLICT (voter_user_id, review_id) DO UPDATE` or a delete, both idempotent, so racing requests converge on the same end state rather than one failing. `insertVote` and `updateVote` collapse into one `upsertVote`. This is also what `docs/schema_docs/planned-database-formats.md` already asked for: *"UI toggle behavior should use idempotent set/remove operations."*

The remove path was already race-safe — deleting a row that is not there is a no-op.

New test races two first votes through a repository double that enforces the primary key the way Postgres does. It fails on 9a23e06 and passes on 5c685e9. Gates: 46 tests, `tsc --noEmit`, `biome check` all pass.

harrietkallberg:
Rebased onto `dev` now that #64 has merged (PR #79), plus one comment fix. Force-pushed as `bc67761`.

**Conflict resolutions** (three files, all additive-vs-additive — both intents kept):

- `meta/_journal.json` — kept both entries in order: #64's `0011_saved_contract` then this PR's `0012_review_contract`. The deliberate `idx: 11` gap is closed now that `0011` is real.
- `server/db/schema.ts` — **both deletions applied**. #64 dropped the `user_favorites` table and its `InsertUserFavorite` / `SelectUserFavorites` types; this PR drops `reviewLikes` and its types. The conflict was between the two survivors, so the resolution was to delete the whole block. Nothing resurrects `user_favorites`, and no `review_likes` reference remains in `schema.ts`.
- `server/api/protected.spec.ts` — kept both coverage sets: #64's saved/taken/collections `it.each` block and this PR's `reviews.vote`.

**Two things a naive resolution would have got wrong**, both fixed:

1. `meta/0012_snapshot.json` was still built against the pre-`0011` schema, so its `prevId` collided with `0011`'s and `db:generate` refused to run at all. Regenerated on top of `0011`. **`bun run db:generate` now reports "No schema changes, nothing to migrate"**, and the migration it would otherwise have emitted is byte-for-byte the hand-written `0012` already in the branch — so `0012_review_contract.sql` itself is unchanged.
2. The journal entry's `when` (`1788477716435`) **predated** `0011`'s (`1788477723351`). `drizzle-orm`'s migrator orders by that timestamp, not by `idx`, so `0012` would have been applied *before* `0011`. It is now `1788480553747`, strictly after.

**Gates on the rebased branch:** `bun run test:web` 90 passed (this PR's and #64's together), `npx tsc --noEmit` clean, `bun run lint` clean. No Neon connection; nothing applied.

**Also in this push** — `0009`'s ownership comment now points `courses_explore_target_sync` at **#77**, which has since been filed ("H: Move search and ingest onto course_explore, then contract the courses table"). It previously said "the not-yet-filed courses/course_explore switch", which was exactly the class of stale pointer this PR set out to fix. Comment lines only; no executable line in an already-applied migration was touched.

**Two coordination notes, neither needing code changes here:**

1. **#65 (PR #78) and #66 will hit a small conflict in `server/errors.ts` and `server/api/trpc.ts`.** This PR adds a `ValidationError` class and two `instanceof` branches. Purely additive, but both files are shared across wave 1. Now called out in the PR description's "Before merging" section so whoever rebases is not surprised.
2. **#77 inherits the `server/search/**` changes made here** — the rewritten `averageRatings` (two surviving 1-10 axes instead of four legacy 1-5 ones) and the star-to-score threshold conversion. It should rebase on this before moving search onto `course_explore`. The PR description says so too.

Issue #102, opened 2026-09-04
DCD2: Schema — derived-data columns, the keyword vocabulary tables, and the schema-doc updates


max-andreasen:
Part of #73. Read the map first — its **Database brief** is the specification for this issue and is authoritative over anything restated here.

Purely additive. This touches nothing #77 holds and must not.

## In scope

**1. Drizzle schema — `apps/web/server/db/schema.ts`**

Five nullable columns on `course_explore`:

| column | type |
|---|---|
| `summary` | `text` |
| `summary_version` | `text` |
| `summary_generated_at` | `timestamptz` |
| `eligibility_version` | `text` |
| `eligibility_extracted_at` | `timestamptz` |

Two new tables:

- `keywords` — `id` serial PK, `term text NOT NULL UNIQUE`, `created_at timestamptz NOT NULL`
- `course_keywords` — `course_code text` FK → `courses.code`, `keyword_id integer` FK → `keywords.id`, composite PK on both, `created_at timestamptz NOT NULL`

**2. An explicit SQL migration** in `apps/web/server/db/drizzle/`.

**3. Update `docs/schema_docs/planned-schema-lucid.json`** — this is canonical and **must land in the same PR**. A migration that ships without it puts the target schema and the database out of sync, which is exactly the failure the schema docs exist to prevent.

**4. Update `docs/schema_docs/planned-database-formats.md`** with two changes that are easy to skip and must not be:

- `course_explore` is currently described as *"derived full-text and semantic-search state."* That is no longer true once a card-rendered summary lives on it. Its charter is now **all derived per-course state**, of which search state is one part. Rewrite that paragraph.
- The document treats the target as **16 tables**. It is now **18**. Change the count wherever it appears and say why.

Add a paragraph covering the new tables: the vocabulary is controlled, zero keywords for a course is a legal outcome, and staleness is tracked by version string rather than content hash because the source prose can no longer change.

**5. `CONTEXT.md`** gains two terms — **Course summary** and **Keyword** — in the existing format, including the `_Avoid_` lines. A keyword is drawn from a controlled vocabulary; that is part of what the word means here and belongs in the definition. Glossary only: no implementation detail.

## Not in scope

- `course_explore.source_hash`. It keeps exactly the meaning #77 gives it. **Do not touch it**
- Any change to `course_prerequisites`. Its provenance columns belong to stage B, which is not designed
[...]

Issue #74, opened 2026-09-03
Spec: search — retrieval input and ranking


max-andreasen:
General spec, jag kikar på en mer detaljerad spec imorgon fredag. 

***

Seed ticket for a structured wayfinder/grilling session. Supersedes the scope of #51, which currently reads "Check the search module / clean up" and is unbounded.

## Scope

Everything about finding a course, split into what goes *into* the index and how hits come *out* of it.

```
D  (no LLM)  A's output + source fields ──► embedding, search_vector
   ranking   two ranked lists ──► one ordered result set
```

## Why search is bad today

Four separate causes, only one of which depends on the other spec:

1. **Embedding input is a wall of prose.** `ingest.ts` embeds `code + titleEng + titleSwe + goals + content` joined into one string. KOPPS `goals`/`content` is long, boilerplate-heavy and structurally near-identical across courses ("After passing the course, the student should be able to…"), so every course's vector looks like every other course's. Biggest single cause. **This is D, and it is the one piece that waits on the other spec.**
2. **`'simple'` tsvector config** (`schema.ts:81`) — no stemming, no stopword removal. "programming" does not match "programs"; "the" is an indexed token.
3. **No score fusion.** `search/service.ts:searchCourses` concatenates: *every* keyword hit first, then unseen semantic hits. A garbage ILIKE substring match outranks a perfect semantic match, always.
4. **ILIKE `%query%` stacked on tsvector** in `searchByKeyword` — two different matching strategies fighting inside one query.

**2, 3 and 4 are unblocked and independently shippable.** They need nothing from the derived-data spec and would likely improve search noticeably on their own. Do not gate them behind it.

## Decisions already taken

- **The seam with the derived-data spec follows the data dependency, not the module boundary.** That spec owns what the LLM produces; this spec owns what is indexed and how it is ranked. The two meet at a contract: *stage D produces an embedding over `<agreed text>` and a `search_vector` over `<agreed text>`, per course, with staleness tracked.* Ranking starts from that as a given.
- **Only D is blocked**, on stage A of the other spec. Not the whole spec.
- **No query-time LLM.** Query rewriting/expansion and LLM reranking put a model call in the latency path of every search. Ruled out.
[...]

Issue #73, opened 2026-09-03
Derived course data — summary, keywords and eligibility extraction


max-andreasen:
> **This issue is now a wayfinder map.** Its previous body was a seed ticket; the decisions it recorded are preserved below under Decisions so far, and the questions it left open were resolved in a charting session on 2026-09-04 (full reasoning in the first comment). Implementation lives in the sub-issues, not here.

## Destination

Every KTH course in Neon carries an LLM-derived **English summary** and a set of **keywords drawn from a controlled vocabulary**, generated by a re-runnable, stage-agnostic derive pass that is decoupled from ingest — with eligibility extraction specified but deliberately held behind reading the real prose first.

Reaching the end of this map means the Course Card stops rendering empty Summary and Keywords sections, and [Spec: search — retrieval input and ranking](https://github.com/kthaisociety/KTH-Course-Community/issues/74) has the stage-A input it is blocked on.

## Notes

- **This map carries execution.** Unlike a normal wayfinder map, most children are implementation tickets meant for AFK agents, not decisions. Two children are decision tickets and are labelled as such.
- **KOPPS is deactivated.** `api.kth.se/api/kopps/v2/*` returns 502 on every endpoint, and KTH have turned it off for good. **Neon is now the source of truth, not a cache.** `courses.goals`, `courses.content` and `courses.eligibility` are irreplaceable — nothing in this map may overwrite them.
- Domain vocabulary: `CONTEXT.md`. Schema target: `docs/schema_docs/planned-schema-lucid.json` (canonical) and `planned-database-formats.md` (prose).
- Server layering is Biome-enforced: router → service → repository → `db`. A new domain means one folder.
- Skills a session should consult: `grilling` and `domain-modeling` for the decision tickets; `tdd` for the implementation ones.

## The pipeline

```
A  (LLM)     goals + content    ──► summary + keywords          ← this map, shipping first
B  (LLM)     eligibility prose  ──► course codes + phrases      ← this map, held behind the corpus read
C  (no LLM)  B's explicit codes ──► course_prerequisites edges  ← this map, codes only
D  (no LLM)  A's output + prose ──► embedding, search_vector    ← #74, not this map
```

Everything reads **raw** prose. There is no stored normalisation step — see Out of scope.

## Database brief

[...]

Issue #110, opened 2026-09-04
Discuss: where should translated and normalised course prose live?


max-andreasen:
Ruled **out of scope** for #73 and filed here for a team discussion. Not blocking anything; #73 ships stage A by reading raw prose and emitting English directly.

## The problem

`courses.goals`, `courses.content` and `courses.eligibility` hold KOPPS prose exactly as it arrived. That prose is:

- **Mixed-language.** Some courses have English syllabi, some Swedish-only. The app is English-only — no locale files, `titleEng` at every render site
- **HTML-laden.** `<p>` tags at minimum; the corpus census under #73 reports the full tag inventory
- **Boilerplate-heavy.** #74 identifies this as the single biggest cause of bad search: *"After passing the course, the student should be able to…"* appears in nearly every course, so every course's embedding vector looks like every other's

## Why #73 deferred it rather than solving it

A normalisation stage — strip HTML, detect language, translate Swedish — sits upstream of everything and is tempting. It was scoped and then dropped, because **nothing currently renders raw prose to a user**. Its only consumers are the derived stages, and stages A and B can read raw prose and emit English directly. Storing normalised prose earns its own call only when something displays it or embeds it.

That second condition is the one likely to arrive: **stage D embeds prose**, and it is #74's, and it currently embeds Swedish HTML.

## What to decide

1. **Is normalised prose stored at all**, or does each consumer normalise on the fly?
2. **If stored, where?** New columns on `course_explore` (whose charter #73 widens to "all derived per-course state"), or its own table. The target grows past 18 tables either way, and `planned-schema-lucid.json` must be updated to match
3. **How much reformatting?** HTML stripping is deterministic and uncontroversial. Translation is an LLM call across the catalogue and is the largest single token cost anyone has proposed here — full `goals` + `content` for ~5k courses dwarfs a 200-character summary by orders of magnitude. Boilerplate removal is a third thing again, and the riskiest: it improves embeddings and destroys meaning if it overreaches
4. **What is the quality bar?** #73 established that nothing renders raw prose today, so "good enough to embed" was sufficient. If normalised prose ever becomes product-facing copy a student reads, the bar and the model both change

## Hard constraint

**KOPPS is deactivated.** `api.kth.se/api/kopps/v2/*` returns 502 on every endpoint and KTH have turned it off. **Neon holds the only copy of this prose that exists.** Whatever is decided, raw prose is never overwritten in place — there is no re-fetch, and an overwrite is permanent loss.

[...]

PR #113, opened 2026-09-04
docs(derive): census the eligibility and syllabus corpus, and commit a fixture (#101)

### LINKED ISSUE #101: DCD1: Dump the eligibility and syllabus corpus from Neon, and census what is in it
Part of #73. Read the map first — its **Database brief** governs this issue and is not repeated in full here.

**This is the corpus read the whole map waits on.** #73's seed body said stage B must not be designed before real eligibility strings have been seen. Nobody has seen them: KOPPS returned 502 through two sessions and there was no `apps/web/.env.local`. Neon now holds the only copy of that prose in existence, and it is reachable.

Nothing here is a decision. This is a fact-finding task whose output unblocks the vocabulary bootstrap, the stage B grilling session, and every cost estimate in the map.

## In scope

**1. A census of what is actually in `courses`.** Query Neon and report, as numbers:

- Total rows, and rows by `state`
- How many have non-empty `goals`, `content`, `eligibility` — separately
- Character-length distribution of each field: min, median, p90, max
- **What fraction of each field is Swedish.** This is the single most important number in the census: it sets the translation burden that the map deliberately deferred, and nobody currently knows it. Use a deterministic detector, not a model
- HTML tag inventory across the three fields — which tags appear, how often. `<p>` is known; find the rest
- How many courses would produce an empty embedding input today

**2. A corpus fixture, committed to the repo.** ~300 courses sampled to span departments and both languages, with `code`, `titleSwe`, `titleEng`, `goals`, `content`, `eligibility` verbatim. This is the reproducible input for every prompt iteration that follows — the point is that prompt work never again depends on database access or KTH's uptime.

Put it under `apps/web/server/derive/__fixtures__/` (create the folder; the derive domain arrives in a later ticket) or `docs/schema_docs/` if that reads better to you. Say which you chose and why.

**3. A short findings note** — `docs/schema_docs/corpus-census.md` — carrying the numbers from (1) and anything surprising. Prose, not a dump.

## Not in scope

- Any schema change, any migration
- Any LLM call. Language detection is deterministic
- Rewriting, translating or cleaning the prose. **Raw prose is now irreplaceable** — read only

## Watch for

Eligibility prose is the reason this ticket exists. Read fifty of them yourself before writing the note, and say plainly in the note whether explicit course codes are common or rare, how requirement phrases are typically worded, and whether the Swedish and English forms differ structurally. That qualitative read is what the stage B session needs; the counts alone will not do it.

## Done when

- `docs/schema_docs/corpus-census.md` exists with every number above
- The fixture is committed and loadable from a test
- The note states, in prose, what eligibility prose actually looks like
- Gates pass: `bun run test:web`, `npx tsc --noEmit`, `bun run lint`

max-andreasen:
Closes #101. Part of #73.

KOPPS is deactivated, so `courses.goals`, `courses.content` and `courses.eligibility` are the only copy of KTH's syllabus prose that still exists. #73 blocked stage B on somebody actually reading it. This is that read — fact-finding only. **No schema change, no migration, no LLM call, and not a byte of the prose is rewritten.**

## What landed

- **`docs/schema_docs/corpus-census.md`** — every number #101 asked for across all 4644 rows, plus a prose account of what eligibility text actually says.
- **`apps/web/server/derive/__fixtures__/corpus.json`** — 300 courses, deterministically sampled across all 100 departments and both languages, stored verbatim (tags and numeric entities untouched). Loaded by `corpus.ts`, asserted by `corpus.spec.ts`.

I put the fixture under `server/derive/__fixtures__/` rather than `docs/schema_docs/` because its job is to be read by code — the ticket requires it to be loadable from a test, and `docs/` holds prose for humans. Reasoning is written up in the note.

## The headline numbers

| | |
|---|---|
| rows | 4644, **all `ESTABLISHED`** — the other two enum values are unused |
| present | `goals` 4374 (94.2%) · `content` 4364 (94.0%) · `eligibility` 3980 (85.7%) |
| **Swedish** | **`goals` 95.7% · `content` 95.3% · `eligibility` 93.6%** |
| plain median length | `goals` 624 · `content` 500 · `eligibility` 109 chars |
| HTML | 7 tags total; **100%** of non-empty values contain markup |
| empty embedding input | 0 literally empty; **254 (5.5%) carry no prose at all** |

## The five things worth your attention

1. **The translation burden is the whole corpus, not a slice.** ~96% Swedish, and 66 courses mix both languages *within one syllabus* — so a per-course language flag would be wrong for some of them. Every stage A prompt is a translation prompt.
2. **97% of rows contain HTML entities, almost all numeric escapes for å/ä/ö** (`&#246;` appears 38851 times in `goals` alone). Decoding them has to happen before anything else — and `course_explore.search_vector` is currently built from the raw text, so the full-text index contains tokens like `246` and `160`. Flagged, not fixed; that is #74's.
3. **Only 33.5% of eligibility rows name a course code at all.** Two thirds of the catalogue will produce zero prerequisite edges, and that is the correct outcome. A plain regex already resolves **84.9%** of the tokens it finds against real rows — the 15% that miss are discontinued courses, not regex noise. Stage C now has a deterministic baseline to beat.
4. **`eller motsvarande` appears in 37% of eligibility rows.** Eligibility is explicitly not a closed set, which is the strongest evidence yet for the map's decision to extract phrases without resolving them.
5. **Absence is `''`, not `NULL`,** in all three columns — against `planned-database-formats.md`'s own convention. Recorded so the next person doesn't write `IS NOT NULL` and get 4644 rows. Not fixed here.

The qualitative read (§"What eligibility prose actually looks like") covers the dominant `Kunskaper i …, 7,5 hp, motsvarande slutförd kurs SF1625` template, the slash-alternative and hyphen-range syntaxes that will trip a naive extractor, and how the English rows differ structurally from the Swedish ones.

## Coordination

Branched from `dev` at `48bb338`, which already contains #77 (merged as #99). Touches nothing #77 or #102 hold — the only new paths are `server/derive/__fixtures__/` and one new doc.

## Gates

`bun run test:web` 183 passed · `tsc --noEmit` clean · `biome check` clean.

https://claude.ai/code/session_0134fagfUVhaUaJXHH31Q9iE


harrietkallberg:
Good job! Please update all the context and docs in this pr before considering merging because they will decide whether the migration updates gets easaily implemented or not. schema docs ->context files -> update schemas -> run migrations. Context drives change, change should not happen before context has approved and defined the target shape. 

DON'T MERGE before frontend has landed on dev :))

harrietkallberg (changes requested):
see my comment, looking good but all the docs need to allign here before merge

harrietkallberg:
@greptile-apps review

harrietkallberg:
## Status: the blocker I put on this is now lifted — here is what is left

Good news first. My review said:

> **DON'T MERGE before frontend has landed on dev :))**

**The frontend has landed.** `85c580f` (#192, the rebuild against the 2026-09-06 design export), plus #194, #197, #198 and #200. That condition is satisfied and is no longer a reason to hold this.

This PR is also **`MERGEABLE`** — unlike its sibling #112, it has no conflicts, because it only adds files (`server/derive/__fixtures__/corpus.{json,ts,spec.ts}` and `docs/schema_docs/corpus-census.md`) and touches nothing `dev` has moved. Greptile is at **5/5** on `8c36b78`, and the frozen-fixture fix answered the shared-cache finding properly.

So the only thing keeping this open is my changes-requested review, and it is worth being precise about what I was asking for.

### What I was asking for

> all the docs need to allign here before merge

Specifically, and it is a smaller ask on this PR than on #112: this ticket's own *Done when* says the note must state **in prose** what eligibility prose actually looks like, and #101's *Watch for* asks you to read fifty eligibility strings yourself and say plainly whether explicit course codes are common or rare, how requirement phrases are typically worded, and whether the Swedish and English forms differ structurally.

That qualitative read is the entire reason this ticket exists, and it is the input **#109 (DCD9) is blocked on** — the stage B grilling session cannot happen without it. The counts alone will not do it. If `corpus-census.md` already carries that section, say so and I will re-read; if it is thin, that is the gap to close.

### Why this matters more than it did on 2026-09-04

This PR is now the **critical path for the whole derived-data map**. #73 is at 0/9 children merged, and two other tickets currently describe this census as not-yet-done and block on it:

- **#109**: *"reading it is a prerequisite nobody has done"*
- **#110**: *"The corpus census … is the factual basis for all four questions. Worth waiting for."*

Both of those are waiting on this PR, not on anyone starting work. I've commented on both to correct the record.

Also worth noting: this PR creates `apps/web/server/derive/__fixtures__/`, so the `derive` folder will exist before #103 builds the domain in it. That is fine, but flag it in #103 so nobody thinks the runner is partly done.

### What it needs to merge

1. **Merge current `dev` in** even though it does not conflict — 16 commits have landed since this was opened, and the gates should run against what is actually there.
2. **Close out the docs ask above**: the prose section on what eligibility text really looks like, and confirmation that the fixture-location choice (`server/derive/__fixtures__/` over `docs/schema_docs/`) is recorded with its reason, as the body asks.
3. **Re-run the gates and report real output**: `bun run test:web`, `npx tsc --noEmit`, `bun run lint`, plus the fixture loading from a test as *Done when* requires.
4. **Fresh 5/5 from Greptile on the final commit** — the current 5/5 is against `8c36b78` and predates the merge. I've re-triggered a review.
5. Then re-request review and I will clear the changes-requested.

One dead premise to note while you are in here: **#101's body says there was no `apps/web/.env.local`.** There is one now. And I re-checked KOPPS today — `api.kth.se/api/kopps/v2/courses.json` and `/course/DD1337` both still return **502**, so the "Neon holds the only copy" framing that makes this ticket urgent is fully intact.

Issue #93, opened 2026-09-04
F9: My Page


harrietkallberg:
Part of #68. Read that issue first — its "Settled decisions", "Conventions" and "Agent workflow" sections govern this one.

## Artboard

`docs/design/Course Community - My Page.dc.html`.

## In scope

The signed-in user's own page, replacing today's `/profile`. Profile basics come from `user.me`; the tabs pull from `saved.list`, `taken.list`, `collections.list` and `reviews.list` filtered to the viewer.

The unreviewed prompt uses the Unreviewed Card from F4 — do not build a second one.

## Profile picture

Multipart does **not** go through tRPC in this repo. `POST /api/user/profile-picture` is the established path and already exists. Do not move it to tRPC and do not add a second upload route.

## Personalization tier

`users.personalization_tier_earned` holds the **highest ever reached**. The effective tier is derived at read time and may decay, and `graph.effectiveTier` returns it. If the page surfaces a tier, render the **effective** one — but never imply the earned value dropped, because it never does.

## Account deletion

`user.delete` exists and cascades. If the design includes it, it needs a confirmation the user cannot hit by accident. Do not add it if the artboard does not show it.

## Done when

- Profile basics, picture upload, and each tab render
- Empty states for each tab are the designed ones
- `ui` tests cover the tabs and at least one empty state
- Gates pass

harrietkallberg:
## Branch model changed — 2026-09-04

**Your PR targets `feat/frontend`, not `dev`.** See the "REVISED: children land on an integration branch" section at the bottom of #68 for the full model.

In short:

- Branch from `origin/feat/frontend`, never from `dev`.
- Open your PR against `feat/frontend`.
- Once Greptile gives you **5/5**, **merge your own PR into `feat/frontend`** — this is the one merge you are permitted.
- **Never push to, merge into, or open a PR against `dev`.** It is protected and requires review; `feat/frontend` deliberately is not.
- Still do not close this issue. Put `Closes #<this issue>` in the PR body for linkage and leave the closing to the product owner.

Greptile reviews your PR on a clean diff against `feat/frontend`, so per-page gating is unchanged — the only difference is that the product owner merges `feat/frontend` into `dev` once at the end instead of merging thirteen PRs.

harrietkallberg:
Merged into `feat/frontend` as PR #126 (Greptile 5/5).

[...]

Issue #91, opened 2026-09-04
F7: Collections


harrietkallberg:
Part of #68. Read that issue first — its "Settled decisions", "Conventions" and "Agent workflow" sections govern this one.

## Artboard

`docs/design/Course Community - Collections.dc.html`.

## Vocabulary — read #68 decision 1 before you start

The artboard says "comparison" in several places. **There is no comparison feature.** Everything is a Collection: `collections.list`, `collections.create`, `collections.rename`, `collections.delete`, `collections.reorder`, `collections.addCourse`, `collections.removeCourse`.

## The rule the database enforces and the UI must respect

**A course may only join a collection if the same user has also saved it.** The composite foreign keys enforce it, and #64's service fails cleanly rather than relying on a constraint violation. So the picker offers saved courses only — do not present an unsaved course as addable and let the server reject it.

## Ordering

`collections.reorder` reorders courses **within** a collection. There is no ordering between collections — `collection_courses.position` is the only ordering the schema carries. It accepts a prefix: members left out keep their relative order behind the ones named, and a non-member is rejected.

## Not-found, not forbidden

Another user's collection reads as a not-found error, never forbidden — ownership is scoped in the query, so the server never learns whether a stranger's id exists. Render it as not-found; do not write copy implying the collection exists but is someone else's.

## Done when

- Create, rename, delete, reorder, add and remove all work
- The picker offers only saved courses
- Empty states for "no collections" and "empty collection" are both designed
- `ui` tests cover the saved-only constraint and a reorder
- Gates pass

harrietkallberg:
## Branch model changed — 2026-09-04

**Your PR targets `feat/frontend`, not `dev`.** See the "REVISED: children land on an integration branch" section at the bottom of #68 for the full model.

In short:

- Branch from `origin/feat/frontend`, never from `dev`.
- Open your PR against `feat/frontend`.
- Once Greptile gives you **5/5**, **merge your own PR into `feat/frontend`** — this is the one merge you are permitted.
- **Never push to, merge into, or open a PR against `dev`.** It is protected and requires review; `feat/frontend` deliberately is not.
- Still do not close this issue. Put `Closes #<this issue>` in the PR body for linkage and leave the closing to the product owner.

Greptile reviews your PR on a clean diff against `feat/frontend`, so per-page gating is unchanged — the only difference is that the product owner merges `feat/frontend` into `dev` once at the end instead of merging thirteen PRs.

[...]

Issue #90, opened 2026-09-04
F6: Saved


harrietkallberg:
Part of #68. Read that issue first — its "Settled decisions", "Conventions" and "Agent workflow" sections govern this one.

## Artboard

`docs/design/Course Community - Saved.dc.html`.

## The rename landed — build against the new names

#64 replaced favorites with the saved domain. Do not build against the old ones:

- `saved.list`, `saved.save`, `saved.unsave` — **there is no toggle**; the old toggle hid which state you ended in
- `user.me` exposes `savedCourseCodes`, not `userFavorites`
- The `user_favorites` table is dropped

The route `/favorites` still exists from before. Move it to `/saved` and update `proxy.ts`'s matcher, which still lists `/favorites/:path*`. Leaving a stale matcher entry means the new route has no signed-out redirect.

## Geometry and the action control

Saved pins the card's `geo` to the **fully collapsed end** — it does not interpolate, unlike Explore. The action control is `action: "add"` ("Add to collection", per #68 decision 1 — the artboard's "Add to comparison" wording does not survive).

## Independence rule

Unsaving removes the save and **must leave taken history and reviews untouched**. The server enforces it; the UI must not imply otherwise — no "this will also remove…" copy, no cascading optimistic updates.

## Done when

- `/saved` lists saved courses and unsaving removes only the save
- `/favorites` no longer exists and `proxy.ts` matches `/saved`
- Empty state is the designed one, not a blank list
- `ui` tests cover the list, the empty state, and that unsave does not touch taken state
- Gates pass

harrietkallberg:
## Branch model changed — 2026-09-04

**Your PR targets `feat/frontend`, not `dev`.** See the "REVISED: children land on an integration branch" section at the bottom of #68 for the full model.

In short:

- Branch from `origin/feat/frontend`, never from `dev`.
- Open your PR against `feat/frontend`.
- Once Greptile gives you **5/5**, **merge your own PR into `feat/frontend`** — this is the one merge you are permitted.
- **Never push to, merge into, or open a PR against `dev`.** It is protected and requires review; `feat/frontend` deliberately is not.
- Still do not close this issue. Put `Closes #<this issue>` in the PR body for linkage and leave the closing to the product owner.

Greptile reviews your PR on a clean diff against `feat/frontend`, so per-page gating is unchanged — the only difference is that the product owner merges `feat/frontend` into `dev` once at the end instead of merging thirteen PRs.

[...]

Issue #89, opened 2026-09-04
F5: Explore


harrietkallberg:
Part of #68. Read that issue first — its "Settled decisions", "Conventions" and "Agent workflow" sections govern this one.

## Artboard

`docs/design/Course Community - Explore.dc.html`. The route is `/search` and **stays that way** — `CONTEXT.md` settles this: Explore is where the user is, search is what they do there.

## In scope

The search-and-browse workspace against `search.courses` and `search.departments`, rendering the Course Card from F2.

**Explore owns the collapse ramp.** It interpolates the card's `geo` object from the results column width. That is the whole reason geometry is a prop — do not push it into the card.

The action control here is the **split Save button** (`action: "save"`).

## Filters, and one thing that changed underneath you

`minRating` now means the **learning-score mean alone, unrounded**, on the 1–10 scale — changed in #67 because averaging workload in meant a punishing course ranked like a rewarding one, which is the opposite of what the filter asks. The dropdown sends a 1–5 star threshold and the server converts it. Keep sending stars; do not send a 1–10 value.

`department` and `minRating` currently post-filter after fetching `size * 5`, which silently truncates. That is a known server-side flaw owned by #74 — **do not fix it here**, but do not build UI that depends on the counts being right either.

## Visitors

Explore is fully open to visitors. Search, browse and reading reviews need no account; saving and taking do. The guest prompts live on the card (F2) — do not gate the page.

## Done when

- Search, department filter and rating filter work end to end
- Cards render at interpolated geometry as the column resizes
- Empty results and the no-query initial state are both designed states, not blanks
- `ui` tests cover the results list, the empty state, and a visitor session
- Gates pass

harrietkallberg:
## Branch model changed — 2026-09-04

**Your PR targets `feat/frontend`, not `dev`.** See the "REVISED: children land on an integration branch" section at the bottom of #68 for the full model.

In short:

- Branch from `origin/feat/frontend`, never from `dev`.
- Open your PR against `feat/frontend`.
- Once Greptile gives you **5/5**, **merge your own PR into `feat/frontend`** — this is the one merge you are permitted.
- **Never push to, merge into, or open a PR against `dev`.** It is protected and requires review; `feat/frontend` deliberately is not.
- Still do not close this issue. Put `Closes #<this issue>` in the PR body for linkage and leave the closing to the product owner.

[...]

PR #116, opened 2026-09-04
feat(courses): the course card Explore, Saved and Collections share

### LINKED ISSUE #86: F2: Course Card
Part of #68. Read that issue first — its "Settled decisions", "Conventions" and "Agent workflow" sections govern this one and are not repeated here.

**The single most reused component in the product**: Explore, Saved and Collections all render it. Get it right here and those three become layout work.

## Artboard

`docs/design/Course Community - Course Card.dc.html`, with `apps/web/data/course-card-sample.ts` as the typed prop shape. **Do not invent a third shape** — the fixture is extracted from the artboard precisely so the card needs no reshaping later.

## Geometry is the parent's job

The artboard passes geometry in as one `geo` object so the **parent owns the collapse ramp**: Explore interpolates it from the results column width, Saved pins it to the fully collapsed end. Keep that split. **Do not fork the card into two components** — the only structural difference between the pages is the action control, which `action: "save" | "add"` already expresses.

## Data

Real, from `dev`: `course.summary`, `course.stats` (batch aggregate), `saved.save` / `saved.unsave`, `taken.add`, `collections.*`.

**Empty for now, per #68 decisions 3 and 4:** keywords, summary, prerequisites. Headers render; sections are empty. Prerequisites need a third state distinct from "None listed".

**Absent, not zero.** A course with no reviews renders the `noStats` / `noReviewStats` states — "No reviews yet" and "Be the first to say how it went." It must never render as 0% happy. `reviews` is empty today, so this is the state you will see constantly; treat it as the primary case, not the edge case.

## Guest states

The artboard has sign-up prompts on both the taken control and the collections picker. Wire them to `AuthReasonDialog` from PR #61 rather than building new ones.

## Done when

- The card renders from `SAMPLE_COURSE` / `SAMPLE_GEO` with no reshaping
- Save, taken and the collections picker work against the real procedures
- `ui` tests cover: no reviews, guest, member, both `action` values, and the collapsed geometry
- Exported from `features/courses/index.ts` for cross-feature use
- Gates pass

harrietkallberg:
Builds the course card from `docs/design/Course Community - Course Card.dc.html` against `apps/web/data/course-card-sample.ts`, the fixture extracted from it. Explore (#89), Saved (#90) and Collections (#91) all render this, so the work here is mostly the seam.

Closes #86

## The seam

Three pieces, each with one job:

| | |
|---|---|
| `CourseCard` | Draws a `CourseCardModel`. Measures nothing, fetches nothing, decides nothing. |
| `toCourseCardModel` | The one place tRPC output becomes that model. Pure, so it is tested without a DOM. |
| `useCourseCard` | Binds one card to the real procedures and owns the picker's transient state. |
| `CourseCardItem` | **What a screen renders.** The keyed boundary the hook needs, and the only one of these the barrel exports. |

A screen writes the loop and nothing else:

```tsx
{courses.map((course) => (
  <CourseCardItem
    key={course.courseCode}
    course={course}
    stats={stats[course.courseCode] ?? NO_STATS}
    geo={geo}
    onOpen={() => open(course.courseCode)}
    onRequestAuth={setAuthReason}
  />
))}
```

`CourseCardItem` is not ceremony. `useCourseCard` holds several hooks, so calling it inside a screen's own `map` callback binds a card's picker and draft to a list *position*: reorder and the open picker jumps to another course, shorten and React throws "Rendered fewer hooks than expected". Greptile reproduced both. Keyed instances make it impossible, which is why the hook is deliberately **not** exported from the barrel.

When #67's aggregation changes, the mapper changes and no card does.

`action: "save" | "add"` is the only structural difference between the pages, so the card is one component rather than two — as #68 required.

## Geometry

`courseCardGeometry(width)` is the artboard's own ramp, lifted from the Explore artboard's interpolation. At full width it lands exactly on `SAMPLE_GEO`, and a test holds those two together — if the ramp drifts, the card is no longer the one the design drew.

The parent still owns it: Explore interpolates from its results column, Saved and Collections use `EXPANDED_CARD_GEOMETRY`. `COLLAPSED_CARD_GEOMETRY` is the other end.

Geometry reaches the card as CSS custom properties rather than inline widths. That is what lets the container query at `@max-[440px]` pin the phone end of the ramp (from the Mobile Preview reference) without the parent knowing a card is narrow — an inline `width` would have outranked it. Mobile is this issue's own, per #68 decision 6.

## Design vs schema — what I hit, and the minimal change made

Per #68's precedence rule, the schema won each of these and the smallest edit that keeps the design intact was made.

| The design assumed | The schema has | Minimal change |
|---|---|---|
| `workload: "3.8"` on a 1-5 scale, bar at 76% | `workload_score BETWEEN 1 AND 10` | Render the mean raw (`"7.6"`); bar width is the mean over ten, which is the *same* 76% the artboard drew. Nothing converts. |
| `"1.2k students have taken this course"`, from KTH enrolment in `cc-store.js` | `user_taken_courses` — app users who marked it themselves | Copy changed to **members**. The number stayed; the sentence was what was wrong. Zero and one are worded so they cannot imply a population that is not there. |
| `hasPrereq` / `noPrereq`, `noPrereq` rendering "None listed" | `course_prerequisites` exists; nothing in `server/` writes it | Third state: both flags false renders an empty row. "None listed" is now reserved for extraction that genuinely found none. |
| `keywords` from a `searchTerms` field | `course_explore` has no such column, on this branch or on `dev` | Header renders over an empty row (#68 decision 3; #73 seeds it). |
| `summary` from `courses.content` | The column exists, but holds KOPPS syllabus prose | Left empty. Rendering it would open every card in a grid with the same boilerplate line. #73 writes the real blurb. |
| The picker adds to a collection independently of Save | `addCourseToCollection` rejects a course the user has not saved | Adding to a collection saves the course first. Flagged here because it is a behaviour change, not just a rendering one. |
| Guest copy promising "compare courses with AI" | No such feature | Deleted, per #68 decision 1. |
| `c.comparisons`, `hasComparisons`, `onNewComparison` | `CONTEXT.md` bans "comparison" in identifiers | Renamed to `collections` / `hasCollections` / `onNewCollection` in the fixture, its types and every use. The button still *reads* "Add to comparison": reader-facing copy follows the design. |
| A fixed amber (`#dfa53c`) for the workload bar | No `--cc-*` carries it; `--cc-warn-btn` is amber only in dark and blue in light, which would make both score bars one colour | Substituted **`--cc-warn-ink`**, the palette's amber-family value in both themes. I briefly added a `--cc-score-workload` token and withdrew it: the palette gap is being settled once at source, not four times by four pages. In dark it sits closer to `--cc-btn` than the artboard intends — **that is the gap, and it is the one thing here still worth fixing in the palette.** |

Two more worth naming:

- **`CONTEXT.md` bans "member" for an app user** while #68's correction mandates "members" in this sentence. `CONTEXT.md`'s own reading rule resolves it — that ban governs identifiers, and this is reader-facing copy — but it is the one place the two documents point in opposite directions.
- **#68 says Saved "pins the ramp to the fully collapsed end"; the Saved artboard does the opposite**, passing a fully expanded geo with a 3-line summary. Both ends are exported, so #90 can take the artboard's. Nothing here forecloses either.

## Other decisions worth a look

- **Prerequisite ticks never cascade.** Marking a course taken writes exactly one row; the tick on a chip means the viewer separately marked *that* course. Covered by a test.
- **The taken pill stops being a control once the course is marked**, rendering as a reading rather than a dead button. The artboard toggles it, but its store holds a boolean flag; ours holds a row with a self-reported grade, credits and periods, and `taken.remove` deletes all of it. Neither the Explore nor the Saved tooltip invites unmarking ("you marked it as taken", "You have taken this course"), so the toggle looks like a mock convenience rather than an intent. Unmarking belongs on Taken courses (#93), where the values at risk are on screen. **Flagging this one specifically** — it is the place I most expect the design owner to overrule me.
- **The whole card opens the course through one control.** The title button carries a stretched `::after` overlay, so a click anywhere opens it while focus and a screen reader meet exactly one affordance — rather than a `role="button"` div wrapping other buttons.
- **Guest prompts are the artboard's inline panels**, but their two buttons hand off to `AuthReasonDialog` from #61. No new sign-in surface, and the screen renders one dialog for its whole list.

## Tests

`ui` project (`features/**/*.spec.tsx`): the card covers no reviews, visitor, signed-in, both `action` values and the collapsed geometry; `useCourseCard` covers what a visitor is asked rather than sent, that marking a course taken writes exactly one row and never a prerequisite's, that the picker saves before it adds, and that a failed write says so instead of drawing a tick that is not there.

The mapper and the ramp are pure, so they go in the `logic` project — including the assertion that the ramp's top end equals `SAMPLE_GEO`, which is what stops the card drifting off the artboard.

`bun run test:web` 297 passed · `npx tsc --noEmit` clean · `bun run lint` clean.

## Greptile

Four rounds. Every in-scope finding is fixed and every inline thread resolved:

- **Partial collection creation left inconsistent state.** Creation and membership are two writes; a failed add was reported as a failed create, which sends the reader back to the name field — and `collections.create` has no uniqueness on name, so retrying made a second empty collection. They now fail apart.
- **The stage-specific error was then lost.** A failed *save* was still being reported as a failed add, pointing at a collection row that cannot work until the course is saved. Writes now throw a `CollectionWriteError` carrying `step`, and the advice follows it.
- **The hook could be called in a list map.** Fixed by construction with `CourseCardItem` rather than by documentation.

The score sits at 4/5 on one objection I have declined: that Explore and Saved still render the legacy card. That is #89 and #90, both `blocked_by` this issue — holding this PR for them is a cycle — and #68 sets the bar as "the app still works", with the shell rollout as explicit precedent for a shared piece landing a round before its consumers. Nothing here changes what those pages render today.

## One documentation fix

`CONTEXT.md`'s **Course prerequisite** entry still carried `_Today_: no table`, but `course_prerequisites` landed in #62/#71, and ADR 0003 has those receipts deleted by the migration that closes them. It is replaced with the fact this card actually depends on: no extracted edges says nothing either way. Without that, the glossary and this diff's comments assert opposite things.

## Notes for the integration branch

`dev` is one commit ahead of `feat/frontend` (`48bb338`, #99 — server-side search state moved to `course_explore`). It touches only `server/**` and does not conflict with this diff. Whether to move `feat/frontend` onto it is the product owner's call, since it changes the base under every open child PR.

Not renamed here: the pre-existing `onAddToComparison` no-op handlers in `course-card-with-charts.tsx`, `search.tsx` and `favorites.tsx`. They are dead callbacks on the components #89 and #90 replace wholesale, and editing them now would only collide with those branches.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -46,8 +46,9 @@ _Avoid_: exam type, assessment method, moment
 **Course prerequisite**:
 A directed edge from a course to one it requires, extracted from the source
 prose. It cannot express AND/OR logic, so it never replaces **Eligibility** —
-extraction narrows, it does not decide.
-_Today_: no table. Closed by `course_prerequisites`.
+extraction narrows, it does not decide. A course with no extracted edges is not
+a course with no prerequisites: nothing in `server/` writes the table yet, so
+absence of edges says nothing either way.
 _Avoid_: requirement, dependency, eligibility
 
 **Eligibility**:
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -46,9 +46,8 @@ _Avoid_: exam type, assessment method, moment
 **Course prerequisite**:
 A directed edge from a course to one it requires, extracted from the source
 prose. It cannot express AND/OR logic, so it never replaces **Eligibility** —
-extraction narrows, it does not decide. A course with no extracted edges is not
-a course with no prerequisites: nothing in `server/` writes the table yet, so
-absence of edges says nothing either way.
+extraction narrows, it does not decide.
+_Today_: no table. Closed by `course_prerequisites`.
 _Avoid_: requirement, dependency, eligibility
 
 **Eligibility**:
--- a/apps/web/data/course-card-sample.ts
+++ b/apps/web/data/course-card-sample.ts
@@ -3,197 +3,48 @@
  *
  * Source: `docs/design/Course Community - Course Card.dc.html` — the
  * `SAMPLE_GEO` / `SAMPLE_COURSE` literals in its trailing
- * `<script type="text/x-dc">` block, and that tag's `data-props` attribute,
- * which names `c: CourseCardModel` and `geo: CardGeometry`.
+ * `<script type="text/x-dc">` block.
  *
- * **Regenerate from the artboard; do not hand-edit.** Values are copied
- * verbatim so a card built against this shape needs no reshaping when the real
- * aggregation lands. Nothing here is invented, renamed or tidied — including
+ * **Regenerate the values from the artboard; do not hand-edit them.** They are
+ * copied verbatim so a card built against this shape needs no reshaping when
+ * the real aggregation lands. Nothing here is invented or tidied — including
  * the things that are plainly wrong (see "Known divergences" below).
  *
+ * The types moved to `@/types/course-card`, which is where the card, the mapper
+ * and every screen read them from; this file is only the artboard's literals.
+ *
  * These are fixtures, not data. The card takes typed props and the *screen*
- * maps tRPC output onto them, so a server change touches one mapping function
- * rather than the card.
+ * maps tRPC output onto them via
+ * `features/courses/lib/course-card-model.ts`, so a server change touches one
+ * mapping function rather than the card.
  *
  * ## Known divergences from the schema — do not "fix" them here
  *
- * - `workload: "3.8"` / `learning: "4.2"` are on the artboard's 1–5 scale.
- *   Scores are **1–10 displayed raw** (issue #68); a 7.6 renders `"7.6"` and
- *   the bar width is `value / 10`, which is the same width the artboard draws.
- * - `comparisons`, `hasComparisons` and `onNewComparison` are the design's word
- *   for **collections**. `CONTEXT.md` bans "comparison" in identifiers and #68
- *   settles the concept as Collection, so the card and everything downstream of
- *   it say `Collection`; these three field names are the one place that word
- *   survives, because #97 requires this file to quote the artboard verbatim.
- *   Every identifier this file coins itself — `CollectionPickerRow` — follows
- *   the glossary instead.
- * - `summary` is `courses.content` in the design's own store, so it does have a
- *   column behind it. `keywords` does not: the store reads a `searchTerms`
- *   field on `course_explore` that the table has no column for, and nothing in
- *   `server/` writes one. Until that is settled the card renders the header
- *   with an empty section (issue #68).
- * - `prereqCourses` entries carry `inCatalog` in the sample but the markup
- *   reads `taken`. Prerequisite ticks are display-only and never cascade.
+ * - `workload: "3.8"` / `learning: "4.2"` are on the artboard's 1-5 scale.
+ *   Scores are **1-10 displayed raw** (#68); a 7.6 renders `"7.6"` and the bar
+ *   width is `value / 10`, which is the same width the artboard draws. The
+ *   mapper does that; these literals predate the scale decision.
+ * - `"1.2k students have taken this course"` is real KTH enrolment in the
+ *   design's mock store. The schema only knows `user_taken_courses` — app users
+ *   who marked the course themselves — so the mapper says **members**, not
+ *   students (#68). The number is real; the artboard's sentence was not.
+ * - `keywords` and `summary` render empty. `course_explore` has no
+ *   `search_terms` column, and the summary maps to `courses.content`, which is
+ *   KOPPS syllabus prose — long, boilerplate-heavy and near-identical across
+ *   courses, so clipping it into a card slot would put the same opening line on
+ *   every card in a grid. #73 generates the real blurb.
+ * - `prereqCourses` entries carry `inCatalog` in the sample but the markup reads
+ *   `taken`. Prerequisite ticks are display-only and never cascade (#68).
+ * - The artboard's `comparisons` / `hasComparisons` / `onNewComparison` are its
+ *   word for **collections**. `CONTEXT.md` bans "comparison" in identifiers and
+ *   #68 settles the concept as Collection, so they are renamed here and
+ *   everywhere downstream. Reader-facing copy still follows the design.
  *
  * `apps/web/data/courseCardMockData.ts` is the older, unrelated mock that
  * existing components still use. This file does not replace it.
  */
 
-/** One prerequisite chip. */
-export interface PrerequisiteCourse {
-  code: string;
-  name: string;
-  /** Whether the prerequisite is itself a course we hold. */
-  inCatalog: boolean;
-  /**
-   * Whether the viewer has separately marked this course taken, which draws
-   * the chip's checkmark. Display-only: marking a course taken never cascades
-   * into its prerequisites (#68).
-   *
-   * The artboard markup reads this; the sample literal does not set it.
-   */
-  taken?: boolean;
-  /**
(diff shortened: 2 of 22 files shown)


harrietkallberg:
## On "Shared card binding is bypassed" — out of scope, deliberately

The finding is accurate as a fact and wrong as a merge blocker, so I am not acting on it. Explore and Saved do still render `CourseCardWithCharts`, and that is the state this PR is supposed to leave them in.

**Wiring those two lists is two other issues, and both are blocked on this one.** #86 is `blocking` #89 (Explore), #90 (Saved) and #91 (Collections) in the native dependency graph. #68 sets the order explicitly — "shell before cards, cards before the pages that render them" — and lists Explore and Saved as later steps with their own artboards. Swapping the list items here would not be an integration touch-up; it would be implementing #89 and #90 blind, without their layouts, against files two sibling agents are about to rewrite.

**The bar for merging is not "the page is done", it is "the app still works."** #68: *"There will be a window where the new Page Header is in while some screens still use the old shell — that is expected and fine, but nothing may be left broken."* That is this situation exactly, one layer down. Nothing here changes what Explore and Saved render today; the card arrives tested and unreferenced, which is what a shared component looks like the round before its consumers land.

Holding this PR until its consumers exist inverts the dependency: #89, #90 and #91 cannot start until it merges.

## What did change since the last round

- Both earlier P1s are fixed and you have verified them.
- The last `onAddToComparison` identifiers are gone from `search.tsx`, `favorites.tsx`, `use-search-page.ts` and `course-card-with-charts.tsx`. #68 decision 1 bans that word in code; the handlers are all `() => {}`, so nothing behaves differently. That is the part of those two files this issue genuinely owns.
- The remove button's hover red now comes from `--cc-danger`, which landed in the palette after this PR opened, instead of shadcn's `--destructive`.

Gates: 296 tests, `tsc` clean, `lint` clean.

harrietkallberg:
## Round 4: the only remaining objection is the one that cannot be resolved here

All three inline P1s are now addressed and their threads resolved:

- **Partial collection creation** — fixed, and verified by you in round 3.
- **Hook order in a list map** — fixed by `CourseCardItem`, and verified by you in round 3 ("a rendered keyed-card list retained AA1000's open picker across reorder and removal with no React hook-order error"). The thread was left open despite that; I have resolved it.
- **Stage-specific error lost after creation** — fixed in `84b028e`. `addToCollection` throws a `CollectionWriteError` carrying `step`, and the create flow branches on it: a failed *add* still points at the collection's row, a failed *save* says to save the course first, because that row cannot succeed until it is saved. Covered by a test.

**The integration objection is the same one raised in round 2 and dropped in round 3.** It is not a defect in this diff, and it cannot be fixed inside it:

- Explore is #89, Saved is #90, Collections is #91. All three are `blocked_by` **this** issue in the dependency graph. Holding this PR until they exist is a cycle.
- #68 fixes the order — *"shell before cards, cards before the pages that render them"* — and sets the merge bar as *"the app still works"*, not *"the page is done"*, with an explicit precedent: *"There will be a window where the new Page Header is in while some screens still use the old shell — that is expected and fine."*
- Rewriting those two lists here means implementing two other issues without their artboards, in files two sibling agents are already working in.

What this PR owns in those files, it has done: the last `onAddToComparison` identifiers are gone (`be7f934`), which is #68 decision 1.

I am treating the integration finding as **answered and out of scope** rather than changing code to satisfy it. Everything actionable within the boundaries of #86 is fixed.

Gates: 297 tests, `tsc` clean, `lint` clean.

Issue #95, opened 2026-09-04
F11: Landing


harrietkallberg:
Part of #68. Read that issue first — its "Settled decisions", "Conventions" and "Agent workflow" sections govern this one.

## Artboard

`docs/design/Course Community - Landing.dc.html`.

## Build it fresh on `dev`

There is a landing page on the `frontend` branch. **Do not cherry-pick it.** That branch predates all of the schema work and is far behind `dev`; #68 is explicit that new screens are cut fresh from `dev`.

`features/landing/lib/hero-geometry.ts` on that branch is useful **reference** for what the client expects, but note its terms differ deliberately: it calls an off-frame node "off-frame", never "anchor", to avoid colliding with the graph domain's word.

## The graph is real now — and currently empty

#65 landed `server/graph/`: `graph.join`, `graph.neighbourhood` (bounded at 150 nodes), `graph.effectiveTier`.

**But nothing calls `graph.join` yet, so no node exists for anyone** — that is #82, unstarted. Until it lands, `graph.neighbourhood` returns not-found for every user. Build "Find your dot" against the real procedure and render a designed empty/unplaced state for that case. Do not fake positions to paper over it, and do not call `graph.join` yourself from the landing page — #82 owns where placement is triggered.

## World units are not pixels

The projection runs **on the client**:

```
screenX = (node.x - currentUser.x) * scale + viewportCenterX
screenY = (node.y - currentUser.y) * scale + viewportCenterY
```

**Never write a projected position back.** Responsive keep-out adjustments are client-side only and must never reach the server. See `docs/landing_docs/personal-community-viewport.md`.

Node appearance comes from the neighbourhood response: `color` is one of six names (`aurora`, `ember`, `frost`, `moss`, `slate`, `violet`) which **the client maps to tokens** — the server stores a name, never a hex value.

## Done when

- The landing page renders for visitors and signed-in users
- Find your dot works against `graph.neighbourhood`, with a designed state for the unplaced case
- No projected coordinate is ever sent to the server
- `ui` tests cover the visitor, placed and unplaced states
- Gates pass

harrietkallberg:
## Branch model changed — 2026-09-04

**Your PR targets `feat/frontend`, not `dev`.** See the "REVISED: children land on an integration branch" section at the bottom of #68 for the full model.

In short:

- Branch from `origin/feat/frontend`, never from `dev`.
- Open your PR against `feat/frontend`.
[...]

Issue #94, opened 2026-09-04
F10: Workspace Pane


harrietkallberg:
Part of #68. Read that issue first — its "Settled decisions", "Conventions" and "Agent workflow" sections govern this one.

## Artboard

`docs/design/Course Community - Workspace Pane.dc.html`.

## Read the artboard before assuming what this is

This is the one page whose purpose is not obvious from its name, and it is not described anywhere in `CONTEXT.md`. **Read the artboard first and state in your PR what you concluded it does.**

If it turns out to introduce a concept the domain language does not have a word for, say so rather than inventing one quietly — that is exactly what `CONTEXT.md` and ADR 0003 exist to prevent. A new concept needs a `CONTEXT.md` entry before it gets identifiers.

If it is a composition of things that already exist — a side pane over Explore, a comparison surface, a detail drawer — build it from the existing components rather than duplicating them.

## Done when

- The pane renders and its purpose is stated plainly in the PR
- Any genuinely new domain concept is flagged for a `CONTEXT.md` entry rather than silently named
- `ui` tests cover its main states
- Gates pass

harrietkallberg:
## Branch model changed — 2026-09-04

**Your PR targets `feat/frontend`, not `dev`.** See the "REVISED: children land on an integration branch" section at the bottom of #68 for the full model.

In short:

- Branch from `origin/feat/frontend`, never from `dev`.
- Open your PR against `feat/frontend`.
- Once Greptile gives you **5/5**, **merge your own PR into `feat/frontend`** — this is the one merge you are permitted.
- **Never push to, merge into, or open a PR against `dev`.** It is protected and requires review; `feat/frontend` deliberately is not.
- Still do not close this issue. Put `Closes #<this issue>` in the PR body for linkage and leave the closing to the product owner.

Greptile reviews your PR on a clean diff against `feat/frontend`, so per-page gating is unchanged — the only difference is that the product owner merges `feat/frontend` into `dev` once at the end instead of merging thirteen PRs.

harrietkallberg:
## Merged as #120 into `feat/frontend` at 5/5. Three things are yours, not the code's.

[...]

Issue #85, opened 2026-09-04
F1: App shell — Page Header and the rail


harrietkallberg:
Part of #68. Read that issue first — its "Settled decisions", "Conventions" and "Agent workflow" sections govern this one and are not repeated here.

**First in the order, because every other screen looks wrong until this is right, and it has no data dependency.**

## Artboard

`docs/design/Course Community - Page Header.dc.html` — authority for layout, spacing and copy.

## In scope

- Replace `AppShell` and `AppSidebar` with the designed Page Header and rail.
- The rail is brand blue in **both** themes (`--cc-rail`), which is deliberate and not a bug.
- Theme toggle wired to `next-themes`, which is already a dependency and already configured in `app/layout.tsx`.
- Signed-in and signed-out states. Visitors may browse; the rail must not imply otherwise.
- Container-query responsive, including the mobile drawer the artboard shows.

## Not in scope

No page bodies. You are replacing the frame every route renders inside, nothing within it.

## The thing to be careful about

This touches every route at once, so it is the child most able to leave `dev` broken. Walk every existing page after the swap — `/`, `/auth`, `/search`, `/course`, `/course/[courseCode]`, `/profile`, `/favorites`, `/reviews`, `/about`, `/contact`, `/newsletter` — and confirm each still renders. A page looking unstyled because its own artboard has not landed yet is expected; a page that throws is not.

`/favorites` still exists as a route while #64 renamed the domain to `saved`. Leave the route alone — a later child owns it.

## Done when

- Every route renders inside the new shell, in light and dark
- `AppShell` / `AppSidebar` are gone, with no dead imports left behind
- Component tests in the `ui` vitest project cover the signed-in and signed-out rail
- Gates pass

harrietkallberg:
## Branch model changed — 2026-09-04

**Your PR targets `feat/frontend`, not `dev`.** See the "REVISED: children land on an integration branch" section at the bottom of #68 for the full model.

In short:

- Branch from `origin/feat/frontend`, never from `dev`.
- Open your PR against `feat/frontend`.
- Once Greptile gives you **5/5**, **merge your own PR into `feat/frontend`** — this is the one merge you are permitted.
- **Never push to, merge into, or open a PR against `dev`.** It is protected and requires review; `feat/frontend` deliberately is not.
- Still do not close this issue. Put `Closes #<this issue>` in the PR body for linkage and leave the closing to the product owner.

[...]

PR #122, opened 2026-09-04
feat(search): Explore, the search-and-browse workspace (#89)

### LINKED ISSUE #89: F5: Explore

harrietkallberg:
Implements the Explore artboard: the search-and-browse workspace at `/search`.

Closes #89

## What is here

- **The page.** `PageHeader` + one 560px search bar + the results column, inside the shell's `PageColumn`. Loading, error, empty and no-query are all drawn states.
- **The collapse ramp, which Explore owns.** `useResultsWidth` watches the results column's own box with a `ResizeObserver` and `courseCardGeometry(width)` turns that into the card's `geo`. Driving it from the *column* rather than the viewport is why the same code will collapse the cards when #94 drags a workspace pane open beside them — Explore never has to learn the pane exists. The card still measures nothing.
- **The split Save button** (`action="save"`), the card's own control for this page.
- **Filters, in the URL.** `department` and `rating`; a filtered search is shareable and Back undoes a filter. The rating dropdown keeps sending **stars** — `search/service.ts` converts to the stored 1-10 scale and thresholds the learning mean alone (#67).
- **Visitors are not gated.** Search, browse and read with no account. The card's own prompts hand off to the single `AuthReasonDialog` this screen renders for its whole list; a visitor clicking Save is asked to sign in, and nothing else is.
- **`?q=` on top of #95's fix.** The seed stays. The debounced query is now written back to `?q=` so the address bar names the results on screen, and — this is the part that needed care — the hook remembers what it wrote, so a `?q=` arriving from **outside** (Back, Forward, a followed link) is adopted into the field instead of being replaced with the stale one. A test moves the URL under a mounted page and holds it to that.

`ui` tests cover the results list, the empty state, the no-query initial state, a visitor session, both filters, the loading and error states, and the URL/field reconciliation. 18 in this file; 434 pass overall, `tsc --noEmit` and `bun run lint` clean.

## The shared-element morph: not built, deliberately

The artboard hands the landing hero's search bar over to Explore — the landing page stashes the bar's rect in `sessionStorage`, Explore translates its own bar out of it, the surroundings fade in behind and the rail slides in on the same curve so the two read as one gesture.

Both halves would have to be written. `features/landing/` stores nothing today, and the rail and topbar belong to `AppShell` — so Explore would be animating chrome it does not own, coupling three merged features for 280ms that the artboard itself already drops under `prefers-reduced-motion` or a stale handoff. The receiving element is the same bar either way, so this can be added later without reshaping anything here. Recorded in `explore.tsx`'s doc block rather than left implicit.

## Design vs schema, and design vs server

Reported per #68's precedence rule. The first four are the server disagreeing with the artboard, not the schema.

| The artboard assumes | What actually exists | The minimal change |
| --- | --- | --- |
| `"12 courses match “x”"` — a count over the whole catalogue | `search.courses` returns one page and reports `total` as that page's own length; `department`/`minRating` post-filter after fetching `size * 5` and silently truncate (#74) | The label says **"Showing N courses for “x”"**. True whatever the server returns |
| A **pager** — `hasPager`, `pageLabel`, prev/next | `search.courses` accepts a `page` input and `searchCourses()` ignores it: there is no offset anywhere | Not built. A pager over one page invents pages that do not exist. For #74 |
| With no query, it lists the whole catalogue (`COURSES.length + " courses"`) | `searchCourses` returns `[]` for an empty query, and no procedure lists the catalogue | A designed starting panel — what to type, plus three searches that run on one click. #89 asks for this state to be designed, not blank |
| A **workspace pane** beside the results, with its own drag handle and floors | #94's, unmerged | Opening a card goes to the course's own page. The ramp already reads the column, so the pane needs no change here |
| The error panel's `#fbeceb` circle and `#a3452a` icon | `cc-theme.css` has no error-surface token; both are light-mode hexes that vanish on the dark page | `color-mix(in srgb, var(--cc-danger) 12%, var(--cc-surface))` over `text-cc-danger`. No token invented |
| No filter row at all | #89 requires department and rating filters | Built in the artboard's own control vocabulary — a 34px pill in `--cc-surface` over `--cc-rule3`. Native `<select>`s: two one-click choices, correct on every platform without a portal |
| Its own sidebar and topbar | The shell (#85) owns both | Explore renders inside `AppShell` and uses `PageColumn` (the 1216px column) and `PageHeader` |

Inherited from #86 and unchanged: `keywords` renders its header over an empty row (`course_explore` has no `search_terms`), and `summary` stays empty (`courses.content` is KOPPS boilerplate — #68's correction 3, and #73's job).

## Two behaviour changes worth naming

- **Explore no longer auto-searches `"interaction programming"`.** That was a hardcoded `DEFAULT_QUERY`; the page now opens on the designed starting state.
- **The `?selected=` details sidebar is gone**, with its hand-rolled resizable `<aside>` and mobile drawer. It is pre-design chrome the artboard replaces with the workspace pane, so keeping it would have left #94 deleting it. `CourseDetailsSidebar` is still exported and untouched.

The surface is named as `CONTEXT.md` names it — `Search` becomes `Explore` — and the route stays `/search`, as #89 requires.


harrietkallberg:
## Disposition of the remaining Greptile findings

**Round 1 — "Delayed filter navigation can overwrite the debounced query URL" (P1): valid, fixed** in `2f56cba`. `router.replace` does not land in `useSearchParams` synchronously, so a school picked while the typed query was still inside its 300ms debounce produced two writes built from the same pre-write snapshot, and the second dropped the first's parameter. A write now composes on the last query string the hook issued for as long as the URL has yet to catch up; any change in the URL clears that record. Two tests hold both orderings, and both fail if the write reads the URL it is about to change. Greptile's round-2 run confirms it resolved.

**Round 2 — "Minimum-rating search can omit matching courses beyond the five-times candidate cap" (P1, ×2): valid as a bug, but deliberately out of scope for this PR.**

Both comments point at `apps/web/server/search/service.ts:133-143`. **This PR touches no server file** — `git diff origin/feat/frontend...HEAD --name-only` is twelve files, all under `app/` and `features/`. The behaviour is pre-existing and unchanged by this branch.

It is also already filed, and this issue is explicitly told not to fix it. #89:

> `department` and `minRating` post-filter after fetching `size * 5`, which silently truncates. That is a known server-side flaw owned by **#74** — **do not fix it here**, but do not build UI that depends on the counts being right either.

Both halves of that instruction are honoured:

- Nothing in `server/` is touched, so #74 keeps a clean problem to solve.
- No UI here depends on the counts being right. The results label reads **"Showing N courses for "x""** rather than the artboard's "N courses match", precisely because the returned set can be shorter than the matching set; and the artboard's pager is **not built**, because `searchCourses()` ignores its `page` input and reports `total` as the length of the page it just returned. Both departures are in the PR description's contradiction table.

Greptile's repro is a good one and worth carrying over — cc #74 for the `size: 1, minRating: 5` case, where five low-rated keyword hits consume the retrieval window ahead of a qualifying sixth.

All four check runs are green, including Greptile Review. Merging into `feat/frontend` on that basis; the outstanding objection belongs to #74 and is recorded here rather than silently dropped.

Issue #88, opened 2026-09-04
F4: Unreviewed Card


harrietkallberg:
Part of #68. Read that issue first — its "Settled decisions", "Conventions" and "Agent workflow" sections govern this one.

## Artboard

`docs/design/Course Community - Unreviewed Card.dc.html`. `docs/design/reference/Course Community - Unreviewed Card Options.dc.html` shows alternates that were **not** chosen — reference only.

## What it is

The prompt shown for a course the viewer has marked taken but has not reviewed. It is the entry point into writing a review, and it is what makes Taken Courses more than a list.

## Data

`taken.list` gives the taken courses; `reviews.list` filtered to the viewer gives what they have already reviewed. The unreviewed set is the difference. Compute it in the screen, not in the card — the card renders one course and knows nothing about the set.

Do not add a server procedure for this. If the difference turns out to be expensive, say so in the PR rather than inventing an endpoint; it is a small per-user set.

## Watch the vocabulary

There is no `happy_took` on a taken course — that field belongs to a **published review** and only exists once one is written. A taken course carries self-reported grade, credits and attendance, plus `transcript_imported_at` provenance. Do not render a satisfaction state for a course that has no review.

## Done when

- A taken-but-unreviewed course renders the prompt and routes into the review flow
- A taken course that already has a review does not appear
- `ui` tests cover both, plus the empty case where everything taken is reviewed
- Gates pass

harrietkallberg:
## Branch model changed — 2026-09-04

**Your PR targets `feat/frontend`, not `dev`.** See the "REVISED: children land on an integration branch" section at the bottom of #68 for the full model.

In short:

- Branch from `origin/feat/frontend`, never from `dev`.
- Open your PR against `feat/frontend`.
- Once Greptile gives you **5/5**, **merge your own PR into `feat/frontend`** — this is the one merge you are permitted.
- **Never push to, merge into, or open a PR against `dev`.** It is protected and requires review; `feat/frontend` deliberately is not.
- Still do not close this issue. Put `Closes #<this issue>` in the PR body for linkage and leave the closing to the product owner.

Greptile reviews your PR on a clean diff against `feat/frontend`, so per-page gating is unchanged — the only difference is that the product owner merges `feat/frontend` into `dev` once at the end instead of merging thirteen PRs.

harrietkallberg:
Merged into `feat/frontend` as PR #117 (Greptile 5/5, 2 rounds).

[...]

Issue #87, opened 2026-09-04
F3: Review Card


harrietkallberg:
Part of #68. Read that issue first — its "Settled decisions", "Conventions" and "Agent workflow" sections govern this one.

## Artboard

`docs/design/Course Community - Review Card.dc.html`. `docs/design/reference/Course Community - Review Card Options.dc.html` shows alternates that were **not** chosen — reference only, do not build from it.

## Data

`reviews.list` and `reviews.byId` are already on target columns after #75. Note what changed there and match the new vocabulary:

- `reviews.vote` takes `{ id, voteType }` where `voteType` is `up` or `down`. The old `reviews.like` is gone.
- **Upvote / downvote, never like / dislike.** `CONTEXT.md` bans the latter outright.
- Downvoting is live — #75 removed the "temporarily unavailable" stub. No dead buttons.
- Scores are **1–10 displayed raw** (#68 decision 2).
- `examination_distribution` is `{exam, assignments, labs, projects, seminars, other}` as whole percentages summing to 100, **or null**. `approach_theory_percent` is 0–100 **or null**. Null means "I don't remember" and must render as an explicit unanswered state, never as zero.

## Editing and deleting

Only the author may edit or delete, and the server already enforces it. The UI must not offer the control to anyone else, but do not treat hiding it as the security boundary.

## Done when

- A review renders with every field, including the null cases
- Voting works both directions and is idempotent on repeat clicks
- Author-only controls appear only for the author
- `ui` tests cover null distribution, null theory percent, both vote directions, and the author/non-author split
- Gates pass

harrietkallberg:
## Branch model changed — 2026-09-04

**Your PR targets `feat/frontend`, not `dev`.** See the "REVISED: children land on an integration branch" section at the bottom of #68 for the full model.

In short:

- Branch from `origin/feat/frontend`, never from `dev`.
- Open your PR against `feat/frontend`.
- Once Greptile gives you **5/5**, **merge your own PR into `feat/frontend`** — this is the one merge you are permitted.
- **Never push to, merge into, or open a PR against `dev`.** It is protected and requires review; `feat/frontend` deliberately is not.
- Still do not close this issue. Put `Closes #<this issue>` in the PR body for linkage and leave the closing to the product owner.

Greptile reviews your PR on a clean diff against `feat/frontend`, so per-page gating is unchanged — the only difference is that the product owner merges `feat/frontend` into `dev` once at the end instead of merging thirteen PRs.

[...]

PR #125, opened 2026-09-04
feat(taken): Taken courses — the list, editing in place, and the Ladok transcript import

### LINKED ISSUE #92: F8: Taken Courses
Part of #68. Read that issue first — its "Settled decisions", "Conventions" and "Agent workflow" sections govern this one.

## Artboard

`docs/design/Course Community - Taken Courses.dc.html`.

## Data

`taken.list`, `taken.add`, `taken.update`, `taken.remove` from #64. Self-reported grade, credits, attendance period and year are all editable.

## Transcript import is real and belongs here

#66 landed it: `POST /api/user/transcript` uploads and parses in one request, returning a **proposal**; `transcript.confirm` writes the accepted rows. This page owns the upload control and the confirm screen.

Points the server work fixed that the UI must not undo:

- **Nothing is written until the user confirms.** Parsing produces a proposal, not rows.
- **Unmatched course codes are reported, never invented.** Show them as unmatched; do not silently drop them and do not offer to create a course.
- **Do not persist or log the file.** It is a student's academic record. It goes up, is parsed, and the text is gone. Do not stash it in component state longer than the confirm step, and never in `localStorage`.
- Imported grades and credits stay **self-reported** and remain editable — the transcript is not authoritative.
- Re-importing does not duplicate: the write is an upsert keyed on `(userId, courseCode)`.

## Provenance

`transcript_imported_at` is the only thing distinguishing an imported row from a manual one, and a manual edit to an imported row **keeps** its provenance. If the design shows a provenance marker, honour that rule; if it does not, do not invent one.

## Done when

- List, add, edit and remove work for manual entries
- Upload produces a proposal, confirm writes it, and unmatched codes are visible
- An unparseable file shows a useful error and writes nothing
- `ui` tests cover the proposal, the unmatched case, and the manual/imported split
- Gates pass

harrietkallberg:
Builds `/taken` from `docs/design/Course Community - Taken Courses.dc.html`: the list of a reader's taken courses, adding one by hand, correcting the self-reported fields in place, removing one, and reading a Ladok transcript.

Closes #92

## What it does

- **The list.** `taken.list` joined with `course.summary` for the titles — `user_taken_courses` stores only a course code, so the lookup is the screen's job and a name that has not arrived falls back to the code. Code, course, credits, grade, year, reviewed, actions, as the artboard's own seven columns.
- **Add by hand.** A catalogue search (`search.courses`) picks the course; credits, grade and year are the reader's own entries.
- **Edit in place.** The pencil opens the artboard's inline editor for credits, grade and year, on every row regardless of where the row came from — a transcript is not authoritative, and `taken.update` preserves `transcript_imported_at`, so an edit keeps its provenance.
- **Remove**, with an undo that puts the whole record back.
- **Transcript import.** `POST /api/user/transcript` reads the file and returns a proposal; nothing is written until "Looks right". A re-read adds the courses that are new and fills the fields that are empty, and never overwrites a correction the reader made by hand — see contradiction 12. Unmatched course codes are named on the confirm screen. An unreadable file gets the artboard's failure screen, which says outright that nothing was saved — because nothing was.
- **The rail's "Taken courses" entry is back**, which #85 deferred only because the route did not exist. `/taken` is titled in the mobile header and added to `proxy.ts`'s matcher.

## The transcript file

It is a parameter to `uploadTranscript` and nothing else. It never becomes component state, never reaches `localStorage`, and is not held by a mutation either — that is why `uploadTranscript` is a plain async function rather than the `useMutation` wrapper the other `api/` files expose: a mutation keeps its last `variables` for as long as its hook is mounted, and the variable here is a student's academic record. What is kept, until the reader confirms or discards it, is the proposal, which holds no file.

## Consumed, not rebuilt

`useUnreviewedTakenCourses`, `UnreviewedCard` (#88), `PageColumn` / `PageHeader` (#85), `useSearchCourses` (#89). `taken.add` still has exactly one write path: `useTakenMutations().add` **is** the course card's `useMarkCourseTaken`, so the two surfaces share one set of cache keys.

## Design vs schema — every contradiction I hit

Per #68's precedence rule: design decides layout, spacing, type, colour, copy, states and responsive; the schema decides what data exists, what it is called, its shape and nullability.

1. **Free-form courses cannot exist.** The artboard's add modal offers `Add "<query>" as a free-form course` — a taken row with no course code — and carries a `match: "freeform"` state through the list, the badges and the reviewer. `user_taken_courses.course_code` is `NOT NULL` with a foreign key to `courses.code`. **Minimal change:** the add form fills code and name from a catalogue search; the free-form path, its explanatory note and the per-row "Free-form course" / "No catalog match" badges are gone. The artboard's own script agrees with the schema here (`"Rows with no catalog match cannot be stored — user_taken_courses has an FK to courses.code"`); only its UI does not.

2. **`taken.update` replaces the whole row.** The artboard's inline editor writes three fields; the procedure's own comment says an omitted field "is the same as clearing it". `attendance_periods` has no column in the design's table, so editing a grade would have silently erased it. `takenUpdateInput` carries the stored value through. Tested in both projects.

3. **The grades switch is not a parse option.** The artboard says grade columns are "left out of the parse, so no grade of yours is stored anywhere". The parse runs on the server and always reads the grade column; there is no request flag and adding one would be server work. **Minimal change:** the copy now says grades are dropped when you confirm, and `planTranscriptImport` strips them from the write. The promise the design actually makes still holds — with the switch off, no grade of the reader's is stored.

4. **10 MB vs 4 MB.** The artboard says "PDF from Ladok · up to 10 MB" in both drop zones. `app/api/user/transcript/route.ts` caps the body at 4 MB and returns 413 above it. The copy follows the server, and the client refuses an oversized file before uploading it.

5. **The "Correct what the parser read" modal has nothing left to correct.** Its fields are code, name and term, and it offers "Revert to the parsed values" against a stored snapshot of what the parse said. Code and name belong to the catalogue, and nothing persists what a parse originally read. Dropped; the row's in-place editor covers the three fields that really are the reader's.

6. **"New" and "Edited" row badges have nothing behind them.** They are session-local flags on the artboard's client state. The durable distinction is `transcript_imported_at` — the one thing that separates an imported row from a manual one — shown as an "Imported" chip on the row and as "Last read 24 Aug 2026" / "Added by hand" in the header.

7. **Undo is offered on hand-entered rows only.** Restoring a removed row goes through `taken.add`, which writes no `transcript_imported_at` and gives no way to set one, so undoing the removal of an imported row would quietly re-label it hand-entered. Rather than lie about provenance, the toast for an imported row carries no Undo.

8. **The proposal has no per-row issues.** The artboard's `checkHeadline` ("N rows need your eyes") counts rows with an `issue` — "No grade in this row of the file" and so on. `TranscriptProposal` reports `candidates` and `unmatched`, nothing else. The confirm screen shows the count and the unmatched list.

9. **The confirm screen lists what it is about to write.** This is the one place I went *past* the artboard rather than trimming it: its `isPreview` screen shows a count, the skipped panel and one button, and no list of the rows. A screen whose single button writes to a reader's academic record has to show what it is about to write, so the candidates are listed in a compact scroller. Everything else on that screen is the artboard's.

10. **The quick reviewer is the review form, not a second one.** The artboard draws a bespoke card stack with its own examination-split dragger, theory slider, two 1–10 sliders, happy-took pair and text box — the same six answers `Review` (#87) already collects, on the same 1–10 scale. Building it would give one review two forms, two validators and two write paths. `UnreviewedCard`'s `onSelect` and `onStart` fill a queue instead and each course opens `Review` in place, which is what #88 built `onSelect` for. Two knock-ons worth naming:
    - The artboard makes the review's text optional ("Optional. You can write the full review later"); `reviewFormSchema({ requireMessage: true })` requires prose for a *first* review. The codebase wins.
    - The artboard's `METHODS` has the design's five examination keys; the schema has six (`seminars`). `Review` already handles six, which is another reason not to draw a second form.

11. **`transcript.confirm` cannot be the whole write.** The design's update dialog promises *"Courses already in your list keep the edits you made — only new rows and missing fields are filled in."* `transcript.confirm` upserts, and its conflict clause sets every self-reported column from the incoming row, `attendance_periods` included — a column `ConfirmedTranscriptRow` has no field for, so it arrives as null. Sending a course the reader already has would replace the credits, grade and year they corrected and erase the periods. **Minimal change, no server work:** `planTranscriptImport` splits the write. A course with no row yet goes through `transcript.confirm`, which is what stamps `transcript_imported_at`; a course the reader already has is touched only to fill a field that is still empty, and then through `taken.update`, which writes the whole row with the stored values and the periods carried, and leaves provenance alone. Found by Greptile; see round 1 on this PR.

12. **Where the rail entry goes.** #85's note said "third in the list, between 'Saved courses' and 'My Page'" in the same breath as calling it "the rail's fourth link". `Explore.dc.html` and `Taken Courses.dc.html` both order the group Explore, Saved courses, My Page, Taken courses. It is fourth.

**No `--cc-*` token was invented.** The artboard's raw hexes for the success banner (`#e9f3ef` / `#1c6b60`), the failure screen (`#fbeceb` / `#a3452a`) and the grade pill (`#fdf7ea` / `#6b4c0d`) are `--cc-success`, `--cc-danger` and `--cc-pill` with token-derived alpha instead.

## Shared code touched

- `Review` gains `triggerless` (renders no trigger button of its own) and its `onEditingClose` is renamed `onClose`, since it already fired on every close and Taken courses needs it to advance its queue. One call site updated.
- `@/features/courses` now exports `useTakenCourses` and `useMarkCourseTaken`; `@/features/search` now exports `useSearchCourses` and `useDebouncedQuery`. Both so a second surface shares one query key rather than wrapping the same procedure again.

## Never touched

`docs/design/` is unchanged.

## Gates

`bun run test:web` (547 passing, 48 files), `npx tsc --noEmit`, `bun run lint` all green. New `ui` tests cover the list, the manual/imported split, add, edit, remove and undo, an upload producing a proposal, confirm writing it, unmatched codes visible, discarding without writing, an unparseable file erroring without writing, the in-place reviewer, a re-read leaving a corrected course alone while filling an empty field, and a refused write keeping the reader's draft.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -55,7 +55,7 @@ apps/web/
 - These layers are **enforced by Biome**, not just convention: a router importing a `repository`, or a repository importing a `service`/`router`, fails `bun run lint`. See the `noRestrictedImports` overrides in `biome.json`.
 - Cross-domain calls go service → service (e.g. `search/service.ts` imports `../course/service`). Routers never import another domain's router; compose them in `server/api/root.ts`.
 - Feature `api/` exposes wrapped `useQuery` / `useMutation` hooks, not raw queryOptions factories.
-- `protectedProcedure` is the real auth gate (`ctx.session.user`). `proxy.ts` (Next 16; not `middleware.ts`) only checks that a session cookie exists for `/profile` and `/saved`. Visitors may browse courses, search, and read reviews.
+- `protectedProcedure` is the real auth gate (`ctx.session.user`). `proxy.ts` (Next 16; not `middleware.ts`) only checks that a session cookie exists for `/profile`, `/saved` and `/taken`. Visitors may browse courses, search, and read reviews.
 - Browser calls same-origin `/api/trpc` and `/api/auth`. Multipart profile pictures POST to `/api/user/profile-picture` (not tRPC).
 - Tests colocate as `*.spec.ts` next to server code; mock repositories, not the database.
 - Domain words: `CONTEXT.md`. Decisions: `docs/adr/NNNN-slug.md`.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -55,7 +55,7 @@ apps/web/
 - These layers are **enforced by Biome**, not just convention: a router importing a `repository`, or a repository importing a `service`/`router`, fails `bun run lint`. See the `noRestrictedImports` overrides in `biome.json`.
 - Cross-domain calls go service → service (e.g. `search/service.ts` imports `../course/service`). Routers never import another domain's router; compose them in `server/api/root.ts`.
 - Feature `api/` exposes wrapped `useQuery` / `useMutation` hooks, not raw queryOptions factories.
-- `protectedProcedure` is the real auth gate (`ctx.session.user`). `proxy.ts` (Next 16; not `middleware.ts`) only checks that a session cookie exists for `/profile` and `/saved`. Visitors may browse courses, search, and read reviews.
+- `protectedProcedure` is the real auth gate (`ctx.session.user`). `proxy.ts` (Next 16; not `middleware.ts`) only checks that a session cookie exists for `/profile`, `/saved` and `/taken`. Visitors may browse courses, search, and read reviews.
 - Browser calls same-origin `/api/trpc` and `/api/auth`. Multipart profile pictures POST to `/api/user/profile-picture` (not tRPC).
 - Tests colocate as `*.spec.ts` next to server code; mock repositories, not the database.
 - Domain words: `CONTEXT.md`. Decisions: `docs/adr/NNNN-slug.md`.
--- a/apps/web/app/(service)/taken/page.tsx
+++ b/apps/web/app/(service)/taken/page.tsx
@@ -1,0 +1,5 @@
+import { TakenCourses } from "@/features/taken/components/taken-courses";
+
+export default function Page() {
+  return <TakenCourses />;
+}
--- a/apps/web/features/courses/index.ts
+++ b/apps/web/features/courses/index.ts
@@ -11,16 +11,23 @@
  * picker and the Collections page are two surfaces over the same seven
  * procedures; exporting the one hook that holds those writes is what stops the
  * second surface from growing a second write path with its own cache keys.
+ *
+ * `useTakenCourses` and `useMarkCourseTaken` are here for the same reason. The
+ * card reads the viewer's taken courses to tick its Taken control and writes
+ * one with `taken.add`; Taken courses (#92) reads the same list and builds its
+ * other three writes on the same invalidation, so both surfaces share one query
+ * key and one `taken.add`.
  */
 
 /** Every `collections` write. Shared by the card's picker and the page. */
-export { useCollectionMutations } from "./api/mutations";
+export { useCollectionMutations, useMarkCourseTaken } from "./api/mutations";
 export {
   type Collection,
   useCollections,
   useCourseDetails,
   useCourseStats,
   useCourseSummaries,
+  useTakenCourses,
 } from "./api/queries";
 /** The presentational card: takes a model, and measures nothing. */
 export { CourseCard } from "./components/course-card";
--- a/apps/web/features/my-page/components/my-page.tsx
+++ b/apps/web/features/my-page/components/my-page.tsx
@@ -409,7 +409,7 @@ export function MyPage() {
           key={editing.review.id}
           courseCode={editing.courseCode}
           editing={editing.review}
-          onEditingClose={() => setEditing(null)}
+          onClose={() => setEditing(null)}
         />
       ) : null}
 
--- a/apps/web/features/reviews/components/review-list.tsx
+++ b/apps/web/features/reviews/components/review-list.tsx
@@ -89,7 +89,7 @@ export function ReviewList({ courseCode, reviews }: Readonly<ReviewListProps>) {
           key={editing.id}
           courseCode={courseCode}
           editing={editing}
-          onEditingClose={() => setEditing(null)}
+          onClose={() => setEditing(null)}
         />
       ) : null}
 
--- a/apps/web/features/reviews/components/review.tsx
+++ b/apps/web/features/reviews/components/review.tsx
@@ -110,15 +110,33 @@ type ReviewProps = {
    * independently — an id belonging to someone else is refused there.
    */
   editing?: EditableReview;
-  /** Required alongside `editing`: the parent owns the dialog's open state. */
-  onEditingClose?: () => void;
+  /**
+   * Renders no trigger button of its own, leaving the opening to `openOnLoad`.
+   *
+   * Taken courses (#92) walks a queue of unreviewed courses and mounts one of
+   * these per course; the row the reader clicked is the trigger, so a second
+   * "Add Review" button sitting under the list would open a dialog that is
+   * already open. `editing` implies this — a review being rewritten is opened
+   * from its own card.
+   */
+  triggerless?: boolean;
+  /**
+   * Called whenever the dialog closes, whether the review was published or
+   * abandoned.
+   *
+   * Required alongside `editing`, where the parent owns the open state, and
+   * alongside `triggerless`, where the parent is what decides whether anything
+   * opens next.
+   */
+  onClose?: () => void;
 };
 
 export function Review({
   courseCode,
   openOnLoad = false,
   editing,
-  onEditingClose,
+  triggerless = false,
+  onClose,
 }: Readonly<ReviewProps>) {
   const { userId, isLoading } = useMe();
   const addReview = useAddReview();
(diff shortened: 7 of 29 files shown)


harrietkallberg:
### Round 2 — both findings taken, and where the remaining edge lives

Both came from the same thing: `confirmProposal` planned against the render's snapshot of `taken.list`. It now refetches the list and plans against that, which answers both at once.

- **Stale snapshot.** A course added in another tab while the proposal sat on screen is now seen and routed to the fill path, so `transcript.confirm` is never asked to create over an existing row.
- **Retries repeating committed writes.** A retry is re-planned against what actually reached the database, so it asks only for what is still missing. The fills also run in sequence now, so a refusal stops instead of firing the rest at a server that has just said no, and the error line says *"Confirming again writes only what is still missing"* rather than claiming nothing was written.

**What this does not do is make the confirm atomic**, and that is deliberate rather than an oversight. Closing the window entirely means `transcript.confirm` leaving existing rows alone — a conditional insert, or the create and the fills in one server-side transaction. That is a change to `server/ingest/transcript` and `server/taken`, which #92 does not own: #68 puts server work outside the frontend track, and #92 states the current contract outright — *"Re-importing does not duplicate — the write upserts on `(userId, courseCode)`."* Overwriting on re-import is the documented server behaviour; the merge on this page is the design's own extra promise layered on top of it, not a correction to it.

So the residual is: between the refetch and the write, a *second* tab could still insert the same course and have it overwritten. It is the same read-then-write window any client has against an unconditional upsert, and it needs a server change to remove. I have not filed it as a separate issue yet — say the word and I will, against `server/ingest/transcript`.

harrietkallberg:
Round 3 fix pushed as `978279d`, on `0cbf9a2` after a rebase.

**The P1 is right, and the harm is larger than the missing message.** `confirmProposal` re-reads the list before planning precisely because the render's copy cannot be trusted at that point — and then fell back to `takenCourses` the moment the re-read failed, which is the one case where that copy is least trustworthy. A refused refresh therefore silently restored the overwrite `0df5198` added the re-read to prevent, and a rejected one escaped as an unhandled rejection leaving the proposal on screen explaining nothing. The refresh and the plan are now inside the guarded block, and a refresh that failed stops the import instead of planning against a stale list.

`isError` rather than the absence of data is what decides that: TanStack keeps the last good rows through a failed refetch, so `current.data` alone cannot tell a fresh list from the stale one.

Two more, found while there:

- **`isConfirming` was only `confirmImport.isPending`,** so "Looks right" went live again between the create and the `taken.update` fills that follow it — a second press mid-import. It is now local state spanning the whole confirm, re-entrancy guard included.
- **`taken.list` had no error branch at all.** A failed read left `rows.length === 0`, which fell through to the *first-run* screen: it told a reader with twenty taken courses that they had none, and offered to import a transcript over a list the page could not see. Same shape as #126's count pill — a derived view is only as loaded as the query under it — so `ListFailed` now says which of the two this is, and offers the read again rather than a drop zone.

Also renamed my-page's `Review` call to `onClose`; this branch renamed the prop from `onEditingClose` before #126 landed a second caller, and the rebase could not see the collision.

`bun run test:web` 584 passed, `tsc --noEmit` clean, `bun run lint` clean.

Resolved the outdated **"Rejected taken-course writes discard the draft"** thread — the round-2 summary disproved it against `236b12e` ("rejected inline edits retained their grade, and rejected manual additions retained the selected course and draft"), it was just never closed.

harrietkallberg:
Round 4 → `bd47e27`.

**"Pending dismissal discards draft" is right and is fixed.** Cancel, the ×, Escape and the overlay all stayed live during `taken.add`, and every one of them ran `reset()` — so a write that then rejected had nothing left to hand back, and the toast pointed at a draft that was already gone. `close()` is now a no-op while `isSaving`, which shuts all four at once: `open` is the parent's state, so Escape and the overlay come back through `onOpenChange` and stop at the same guard. That is the guard `TakenCourseRow` already had on its own Cancel; the dialog was the outlier. Covered by *"cannot be dismissed out from under a write still in flight"*.

**On the second item in the summary — the transcript confirmation race — I am not going to close it here, and I want to be explicit about why rather than leave it looking unaddressed.**

It is real, and it is the residue of the fix in `0df5198`: `confirmProposal` re-reads `taken.list` and plans against that, but a course added from another session *after* the refetch resolves and *before* `transcript.confirm` lands is still planned as a create, and that upsert's conflict clause sets every self-reported column from the incoming row — clearing `attendance_periods`, which the payload has no field for.

**The window cannot be closed from this page.** Nothing between the read and the write is one transaction, and no amount of re-reading changes that; the check and the write are two round trips by construction. Closing it means `transcript.confirm` inserting conditionally and routing conflicts through a missing-field-only update — a change to the server contract #66 landed and #92 explicitly does not own. #92 states the contract it inherits outright: *"the write is an upsert keyed on (userId, courseCode)"*. Rewriting that under an F-series UI issue would put a schema-adjacent behaviour change in a PR nobody is reviewing for it.

So it is filed rather than fixed, and the window is now as narrow as a client can make it. Filing it as a follow-up on `dev` next, and linking it here.

`bun run test:web` 585 passed, `tsc --noEmit` clean, `bun run lint` clean.

harrietkallberg:
Closes #92

PR #199, opened 2026-09-06
feat(taken): open Taken courses to signed-out visitors, gating at the keep step

harrietkallberg:
Opens `/taken` to signed-out visitors, the last page still redirecting them to `/auth`. Stacked on #194, which did the same job for `/saved` and `/profile` — retarget to `dev` when that merges.

## What the artboard asks for

`docs/design_ref/2026-09-06/Course Community - Taken Courses.dc.html`:

- **:767** — `if (h.isGuest) return { ...base, signedIn: false, screen: "empty", … }`. A guest lands on the drop zone, not a locked page.
- **:1305-1308** — the gate, and the point of the whole thing:
  ```js
  confirmCta: s.signedIn ? "Looks right" : "Sign in to keep this list",
  onConfirm: () => {
    if (!s.signedIn) { this.setState({ screen: "auth", returnTo: "list", pending: "confirm" }); return; }
    this.saveList();
  }
  ```
- **:41-45** — the rail's guest banner, which `rail.tsx` already renders and could never reach this page.

## What changed

**Both gates are gone.** `useRequireSession()` came out of `taken-courses.tsx`, and `/taken` came out of `proxy.ts`'s matcher — which left the matcher empty, so **`proxy.ts` is deleted**. `protectedProcedure` is now the only auth gate in the app, which is what ADR 0001 and 0002 always called the real one; the proxy was the optimistic cookie check standing in front of it. `CLAUDE.md`/`AGENTS.md` said it gated three routes and now says there is no proxy.

**`POST /api/user/transcript` answers a signed-out caller.** Safe because parsing and writing are two calls, which the route header has said since #66: it reads the file, returns a proposal and stores nothing. **`transcript.confirm` stays `protectedProcedure` and is untouched** — nothing a signed-out caller does writes a row.

**The proposal is handed across the sign-in.** `features/taken/lib/guest-proposal.ts` — never the file, grades stripped unless the reader turned the switch on, thirty-minute expiry, cleared the moment an account claims it. Same answer `workspace-storage.ts` gives for an unpublished review draft, for the same reason: the magic link opens in a new tab, so React state and `sessionStorage` are both empty by construction.

## The unauthenticated parse, and its bounds

Opening the route means running a PDF parser for anyone who asks. Bounded before it spends any CPU:

| Bound | Where | Holds against |
|---|---|---|
| 4MB body cap, counted as it streams | `capRequestBody` (unchanged) | anything |
| 2 concurrent parses, 8 queued, 5s wall-clock, `terminate()` | `pdf-text.ts` (#121, unchanged) | anything |
| 1 signed-out parse in flight, process-wide | **new** | a signed-out flood starving accounts |
| 5 signed-out reads / 15 min per caller | **new** | accidents, retry loops, casual abuse |
| 120 signed-out reads / 15 min, all callers | **new** | the same flood across forged headers |

The limits are checked **before** the body is read, so a refused caller never gets 4MB buffered on their behalf. A signed-in reader is subject to none of them.

**The honest caveats**, written into `server/http/rate-limit.ts` rather than glossed:

- The counters are **in this process's memory**. The app ships `output: "standalone"` in a container, so they survive between requests — but two replicas is two ceilings, and a redeploy resets them.
- The key comes off `x-forwarded-for` / `x-real-ip`, which is **client-supplied**. Anyone willing to vary that header gets as many buckets as they like.
- So the per-caller limit is a **speed bump**. What holds against a determined attacker is the row above it: the concurrency cap bounds the CPU regardless of request rate, and the new in-flight gate keeps a parser slot reachable by an account at all times.
- The real fix for the distributed case is a shared counter. There is no Redis in this stack and one route does not justify adding one; `rate-limit.ts` is the seam to move behind when one arrives.

## Judgement calls, stated

- **The resume stops one step short of the artboard.** It confirms the moment the account appears; this puts the rows back on the preview under "Looks right" and the reader clicks. `use-guest-saves.ts` settled the same question the same way — *signing in is not consent to write a list of courses to an account* — and it matters more here, because the rows are a student's record and because `localStorage` is somewhere else on this origin can write.
- **The sign-in is `AuthReasonDialog`, not the artboard's `isAuth` screen.** That is the sign-in surface everywhere else in this app, and it already carries the `callbackURL` and email hand-off the artboard's screen only mimes. The new `keep-course-list` reason uses the artboard's kicker and title word for word (`:58-59`).
- **"Add courses manually instead" is gated too.** The artboard's handler just opens the form — its mock has no server to refuse the save. `taken.add` is protected, so a guest gets the prompt instead of a form whose Save cannot work.
- **OAuth loses the in-memory proposal but not the stored one.** `signIn.social` navigates away; the record is what brings the rows back, which is why it exists.

## Verified

- `bun run lint`, `bunx tsc --noEmit`, `bun run test:web` — clean (8 pre-existing warnings in `.agents/skills/`). 1271 tests, three consecutive green runs.
- **Every new gate test was seen to fail with the gate put back.** Restoring `useRequireSession()` and the confirm gate fails 4; restoring the manual-add gate and the resume effect fails 3; restoring the 401 fails 4 route tests.
- **Run against the real app.** A signed-out visitor reaches `/taken` and gets the artboard's empty screen — Resultatintyg hint, grades toggle, drop zone, "Add courses manually instead", rail guest banner. "Add courses manually instead" raises the prompt. Six unauthenticated `POST`s to the route: five reach the parser (422 on a junk PDF), the sixth is `429` with `Retry-After: 890`.

## Left undone

- `docs/adr/0001` and `0002` mention `proxy.ts` in the present tense. Left alone — an ADR records a decision at its date, and rewriting one to match later code is worse than a stale reference.
- `audit-ledger.md`'s Route 5 cites `docs/design_ref_new/`, a dead path. Its paths are **not** rewritten, because the artboards are revised at each export and a rewritten path would make claims about a superseded revision read as current. A note says so, and records the one exception this PR adds to T-04's "never in `localStorage`".

https://claude.ai/code/session_01RAUnEjgKvSqyiJxi3GWbWg


harrietkallberg:
@greptile-apps review

Issue #191, opened 2026-09-06
21% of the code is comment, and every restatement is a source of truth that can drift


harrietkallberg:
The reconciliation added a great deal of prose to the codebase. Re-measured on
`dev` @ `4fda3d3`, across 144 non-spec files of 80 lines or more:

**7,834 comment lines out of 34,071 — 23% of the code is comment.**

The original measurement was 6,731/32,171 (21%) across 135 files on
`feat/frontend`. **The ratio has risen since this issue was filed, not fallen**
— nine more files crossed 80 lines and the comment share went up with them. The
trend this issue describes is still running.

The twelve heaviest:

```
65%   59/91    features/my-page/api/queries.ts
62%   96/156   features/workspace/lib/review-draft.ts
57%   57/100   features/auth/lib/return-to.ts
57%  246/432   features/landing/lib/neighbourhood-view.ts
57%  212/374   features/landing/lib/hero-keepout.ts
56%  173/308   features/saved/lib/guest-saves.ts
56%   89/160   server/graph/tier.ts
54%   80/149   components/ui/confirm-dialog.tsx
52%   85/162   features/reviews/lib/reviewer-session.ts
51%   76/150   features/saved/hooks/use-guest-saves.ts
51%  128/253   features/shell/components/search-morph.tsx
50%  233/463   features/reviews/lib/review-draft.ts
```

Four files dropped out of the top twelve **because new ones displaced them, not
because they improved.** They are unchanged and still in scope:

```
50%  103/208   server/graph/appearance.ts
49%   51/105   features/workspace/hooks/use-workspace-pane.ts
48%  289/598   server/graph/service.ts
48%   64/134   features/saved/api/mutations.ts
```

Two corrections to the original table: `hero-keepout.ts` grew 206 → 374 lines
(the #187 radial-push fix), and the path `server/saved/api/mutations.ts` never
existed — it is `features/saved/api/mutations.ts`.

The three new entrants — `my-page/api/queries.ts`, `saved/lib/guest-saves.ts`
and `saved/hooks/use-guest-saves.ts` — arrived with the guest-page work (#194,
#200), which postdates this issue. That is the mechanism described under "How it
happened" still operating.

## Why this is a defect and not a style preference

**Every restatement is a source of truth that can drift.** A comment explaining
what `CONTEXT.md` already defines, or what the schema already states, is a second
copy — and when the first copy changes, nothing makes the second follow. This
[...]

Issue #190, opened 2026-09-06
Index: the #167/#168/#169 leftovers, split into #214-#219


harrietkallberg:
**Split up and closed on 2026-09-07.** Nothing here was dropped — every item that was still live has moved to its own issue, listed below. This stays as the index that explains where each piece went.

This issue collected everything left over from the #167/#168/#169 audit that PR #187 could not do. It carried `ready-for-agent`, but only three of its six remaining items were actually specified well enough to hand to an agent cold: two needed a design pass first, and three were decisions nobody had made. Six unrelated items behind one label meant an agent grabbing it could not tell what it owned, so the whole thing sat still for a month. Hence the split.

## Ready to pick up now

| | |
|---|---|
| #214 | Delete the dead `focus-visible:outline-cc-brand` on the Explore school filter *(was item 3)* |
| #216 | Extract the duplicated row-to-`TakenCourseInput` mapping in `confirmTranscriptImport` *(was item 2, first half)* |

Each is self-contained, names the trap that makes it non-obvious, and states what "done" means.

#215 (record why the review card omits the artboard's term) was also filed from item 5's second bullet and **closed as not needed**: the artboard declares `term` in its own prop list and draws it, but `cc-store.js` supplies no `term` on a review, so the divergence is internal to the design export rather than between design and code. `review-card.tsx` has nothing to explain.

## Needs a design pass before it is agent work

| | |
|---|---|
| #217 | One owner for the spend-once `?open=` guard, duplicated across Explore and Saved *(was item 1)* |
| #218 | The transcript import's always-zero `updated` count, and the UPDATE-per-row fill *(was item 2, second half)* |

Both describe a real duplication with a fix shape that is **not** obvious — #217 because the two copies clear the URL by different mechanisms and one must preserve `?collection=`, #218 because de-looping needs a batch repository method that does not exist. The open questions are listed on each; answer them and they become normal tickets.

## Needs a decision

| | |
|---|---|
| #219 | The hero push's last resort, the 39 unused primitives, and the magic-link metrics *(was items 4, 5-first-bullet and 6)* |

Filed as questions rather than as work, because filing them as work is what has already failed twice.

## Closed while this issue was open

Recorded because #187, #194 and #200 cross-reference these, and a reader following one of those links should land on an answer rather than a gap.

[...]

Issue #161, opened 2026-09-05
The design contradicts itself about which tier buys which personalization axis


harrietkallberg:
My Page's "My dot" tab lists three personalization axes and says which tier
unlocks each. The design states two different answers, and the revision did not
settle it, so the code had to pick one without authority.

## The contradiction

From `apps/web/features/my-page/lib/personalization-tiers.ts:33-42`:

- The My Page artboard's **rendered list** builds `mk(2, "Dot style", ..., "style")`
  and `mk(3, "Signal on click", ..., "signalStyle")`.
- `cc-store.js`'s **`TIER_AXES` constant** is
  `{ 1: "color", 2: "signalStyle", 3: "style" }` — the opposite pairing for
  tiers 2 and 3.

Both halves survive unchanged into the current `docs/design_ref_new/` export, so
this is not stale: the revised design still disagrees with itself.

## What the code does, and why it is provisional

It follows the **rendered list**, on the grounds that this is what a reader of
the artboard actually sees. The schema cannot break the tie:
`users.personalization_tier_earned` is one number, and no column says which axis
a tier buys.

This is the normal precedence rule failing, because the rule is "match the
artboard" and the artboard says both things. The choice made in the code should
be treated as provisional rather than as design conformance until someone
decides.

## Related, and not blocking

Nothing in `server/` writes `personalization_tier_earned`, so every account is
at tier 0 and all three axes render as locked. That is handled honestly — the
`unlockHint` copy describes the intended rules rather than a mechanism that
runs, and `node-profile.tsx:33-47` says so. The revised `cc-store.js` agrees:
"Tier 0 accounts keep the default look; only personalized nodes carry a row."

So this decision is not urgent — no reader can currently reach tier 2 or 3. It
should be settled before the column gets a writer, because at that point the
pairing becomes visible and changing it would move a feature a member had
already unlocked.

Found by the #134 cross-route audit; recorded as **M-04** in `audit-ledger.md`.

harrietkallberg:
# Settled: what earns each personalization tier — 2026-09-05

The contradiction this issue reports is resolved by deciding the ladder, which
nothing had ever done. That is why `personalization_tier_earned` has no writer:
not an oversight, an unmade decision.

## The ladder

**Tier 1 — the app user has published a review.**
The 0→1 step, and the single contribution the product most wants. It also lines
up with machinery that already exists: `deriveEffectiveTier` measures decay from
[...]

PR #202, opened 2026-09-06
docs(comments): audit the comments against #191's rubric

harrietkallberg:
Closes part of #191 — the comment audit itself. 93 files, comments only.

**This is not a ratio result and the number is not the point.** Density across `apps/web` went 23.1% → 22.9%. Several of the heaviest files are *unchanged or barely changed* because they are heavy for good reasons, and one file got *longer*. What moved is the number of comments asserting a fact that nothing checks.

## Comments-only: how it was verified

Every changed file was run through the TypeScript compiler with `removeComments: true` (CSS through an exact block-comment scan), the output whitespace-collapsed, and the result compared between `dev` and this branch. **93 of 93 files produce byte-identical code.** The checker was itself sanity-checked by appending `export const __probe = 1;` to a file and confirming it reported `CODE CHANGED`.

### 1. Stale facts — comments that were already wrong (8 files)

The expensive kind. Each was verified against the code or the current artboard before being touched.

| Comment | Reality |
| --- | --- |
| `globals.css`: "the artboards carry 22 inline `outline:none` declarations" | The 2026-09-06 export carries **25**. Same failure mode as the `globals.css` count named in the #191 thread. |
| `explore.tsx`: "`search.courses` … reports `total` as the length of that page" | There is no `total` on that reply at all. The comment described a shape the code had left behind. |
| `my-page.tsx`: "no longer redirected away by a proxy … the stale-cookie case that used to be the only way to reach this branch" | There is no `proxy.ts`. Nothing gates the route. (It also contained a literal `no no`.) |
| `course-card-item.tsx` + `features/courses/index.ts`: "Saved and Collections pin an end" of the card ramp | Neither does. `saved.tsx` computes `courseCardGeometry(resultsWidth)`; `collection-detail.tsx` takes a measured `geo`. `card-geometry.ts` already warned that naming screens is how these drift — and they had. |
| `reviewer-session.ts`: "the round-level refusals **below**" | They are above. |
| `saved.tsx`, `globals.css` ×2: cited the **2026-09-05** export | Re-read against 2026-09-06. All three claims still hold; re-cited by artboard name. |
| `collections.tsx:524`/`:553` cited from `globals.css` | Already nine lines out. |

### 2. Artboard and source line citations — 40 removed, 0 remain

`#191` names these specifically: they have gone stale at *every* export, and because artboards are revised rather than moved, a stale one is wrong about the design and not merely about the path. Every `.dc.html:NNN`, `(line NNN)` and bare `` `:NNN` `` is gone from `apps/web`. The artboard name stays, and so does the identifier inside it where that was doing the work — `toExplore()`, `pageLabel`, `savedLabel`, `pendingImport`, `searchBarMargin`, `runImport`.

Before: `` (`Course Community - Explore.dc.html:264`, whose `pageLabel` is built at :1289) ``
After: `` (its `pageLabel`) ``

### 3. Issue numbers used as explanation — 184 → 52

An issue number appended to a standing statement says which PR made the code this shape, not why the shape is right. 54 bare `(#NNN)` attributions were stripped mechanically; the rest were rewritten sentence by sentence.

**`#68` is kept everywhere — all 47 of them.** The binding frontend product calls live in that issue's comments, so `#68 §5` points at a live decision, not at history. `#87` is kept in `examination-palette.ts` for the same reason.

Where a reference stood in for a decision that has since been written down, it now points at the decision: `docs/adr/0005-personalization-tier-earning.md` replaces `#161` in `tier.ts`, `reviews/service.ts`, `ingest/transcript/service.ts` and `graph/repository.ts`.

The 52 that remain are almost all in `*.spec.*` files, where the number identifies the regression the test pins. Those are provenance for an assertion rather than explanation of a shape, and #191's rubric is about the latter.

### 4. History narration (≈30 passages)

Before: *"This file used to hold a byte-identical copy of `useTakenCourses`, and that was not a tidiness problem: …"*
After: *"…re-exported rather than re-declared: a second declaration is a second `useQuery` call site with its own cache entry, so My Page and the course card would fetch `taken.list` twice and could disagree about the answer."*

The consistent move is **from anecdote to rule**, because a rule survives the next change and an anecdote does not:

- `review-draft.ts`: "Both old copies started `{ ...EMPTY_REVIEW_DRAFT, ... }`…" → "Do not spread it, and here is what a spread costs."
- `dialog.tsx`: "There used to be an `sm:max-w-sm` beside it, and it was a trap" → "**Do not add a responsive clamp beside it**", with the tailwind-merge mechanics intact.
- `review.tsx`: "It used to be `max-w-4xl min-w-3xl`, and the floor was the bug" → "Do not add a `min-w-*` floor beside it", with the CSS precedence and the phone symptom intact.
- `reviewer-session.ts`, `guest-saves.ts`, `use-guest-saves.ts`, `guest-proposal.ts`, `open-courses.ts`, `use-explore.ts`, `saved.tsx`, `workspace-storage.ts`: same treatment.

### 5. Refactor predictions (#190 item 8)

`features/collections/index.ts` predicted that a third screen needing `ConfirmDialog` should move it to `components/ui/`. A third screen arrived and built a near-verbatim copy anyway. The move has since happened — and the prediction, *plus the story of it being missed*, had been restated in four files. All four now state the rule that still binds (the shell is one component; do not build a second) and none tells the story.

**Two forward-looking comments were removed rather than rewritten. They should become issues:**

1. `components/ui/confirm-dialog.tsx` — "`AlertDialogContent` still carries the `sm:max-w-sm` clamp; removing it is a change to a shared primitive that belongs in its own PR". The *trap* is kept and stated at both `confirm-dialog.tsx` and `alert-dialog.tsx`; only the promise of a future PR is gone.
2. `components/ui/alert-dialog.tsx` — "which vendored shadcn primitives should this repo still carry is one question about 37 files … see the follow-up issue". No such issue is findable from the comment. The file's real, checkable fact — it has no consumer — is kept.

### 6. The same explanation in several files

The `--cc-danger-tint`-not-`color-mix` argument appeared in full in `collection-chip.tsx`, `collection-tile.tsx`, `explore.tsx` and `feedback-form.tsx`. It now lives once, in the `globals.css` panel-tint block; the call sites state the rule in a line and point there. Same for the removed rating filter, whose lesson (a filter this domain cannot express in SQL breaks paging) is stated once in `server/search/service.ts`.

## What was deliberately kept

**The three files #191 defends.** Cuts in them are small and each is defensible:

- **`features/landing/lib/hero-keepout.ts`** — one paragraph, the retired synthetic-dot field. Everything else stays, including the closest-pair measurements (3.35px → 0.35px), the `(400, 235.999999)` collapse, the `PUSH_EPSILON` floating-point argument, and the "known residue, do not read the radial path as a guarantee" warning.
- **`features/landing/lib/neighbourhood-view.ts`** — the scale/anchor history compressed into the proof that survives it (`screen(p) − screen(q) = (p − q) · s`); a commit SHA dropped from the `ANCHOR_COMFORT` note while its 1920×600 / 360×480 measurements stay. The `VIEW_SCALE` density argument, the "no density policy" statement and the #68 supersession are untouched.
- **`server/graph/tier.ts`** — three edits, all citation swaps to ADR 0005. The ladder, the highest-rung-wins reasoning and the non-monotonicity paragraph are untouched. The ladder table is technically a second copy of ADR 0005, and it stays: it is the rule a reader of `deriveEarnedTier` needs at the call site, and ADR 0005 itself says the asymmetry is written at the rule, the writer and the repository on purpose.

**Traps, kept and strengthened.** `review-draft-panel.tsx`'s `peer-focus-visible` ring and `workspace-pane.tsx`'s `focus:bg-cc-pill` menu highlight both now open **"Do not remove this as a duplicate"** instead of citing a sweep's PR number — so the warning no longer depends on knowing what that PR was.

**External constraints, untouched.** The HNSW approximation caveat and the total-order prefix-stability argument in `server/search/`; the tailwind-merge group split; `schema.ts`'s `ALTER TYPE … ADD VALUE` note.

**Whole files left alone** because every line earns its place: `features/saved/api/mutations.ts` (the unsave→`collection_courses` cascade is invisible from the client and nothing but a comment can say so), `features/shell/components/search-morph.tsx` (spring constants derived from ζ and ω), `features/workspace/hooks/use-workspace-pane.ts` (why `hydrated` is state and `read` is a ref — opposite answers, and swapping them is the bug at both ends), `server/http/rate-limit.ts` and `app/api/user/transcript/route.ts`, `features/taken/lib/guest-proposal.ts`'s capability argument.

## Left alone pending an unanswered product question

**Which personalization axis each tier unlocks.** `server/graph/appearance.ts` says the pairing "was settled by the product owner on 2026-09-05: the rendered My Page artboard wins". ADR 0005 says the axis half of #161 is **still open** and that the code follows the rendered list *provisionally*. Both cannot be current. Every factual claim in that comment was re-verified against the 2026-09-06 export and all of them hold — the artboard draws `mk(1, "Dot color")`, `mk(2, "Dot style")`, `mk(3, "Signal on click")`, and `cc-store.js` still says `TIER_AXES = { 1: "color", 2: "signalStyle", 3: "style" }` — so the comment is accurate about the artboards and possibly wrong about the status. Resolving that is a product call, so **the comment is untouched**. ADR 0005 flags this as urgent now that members can actually reach tiers 1–3.

## Out of scope, reported not fixed

1. **`useRequireSession` has no non-test consumer.** `features/auth/hooks/session.ts` exports it, `features/auth/index.ts` re-exports it, and the only references left in `apps/web` are two spec files mocking it. It is the last of the client-side route gating. Dead code.
2. **ADR 0003 has gone out of date about "guest".** It states that *"the code already respects this line — 'guest' and 'member' appear only in copy, never as an identifier"*. That is no longer true: `useGuestSaves`, `toggleGuestSave`, `GuestImportState`, `GUEST_SAVE_PREFIX`, `guestReads`, `guestParses`, `guest-saves.ts` and `guest-proposal.ts` are all identifiers, and `CONTEXT.md` lists "guest" under **Visitor**'s `_Avoid_`. Either the glossary or the identifiers should move.
3. **`components/ui/chart.tsx` and `components/ui/alert-dialog.tsx` have no consumers.** Both are noted in comments; neither is imported anywhere in `apps/web`.
4. **Pre-existing `TODO`s in `apps/web/types/course.ts`** (`examTypes` "currently just a placeholder", "consider if we actually need both department and code") predate the reconciliation and are forward-looking in the way #190 item 8 warns about. Left alone under "when in doubt, keep" — they are honest markers of known gaps rather than predictions of a refactor, but they would be better as issues.

## Checks

```
$ bun run lint
Checked 427 files in 627ms. No fixes applied.
Found 8 warnings.        # exit 0
```

All 8 warnings are in `.agents/skills/neon-drizzle/` (`generate-schema.ts`, `run-migration.ts`, `db-http.ts`, `db-websocket.ts`, `drizzle-config.ts`, `schema-example.ts`) — files this branch does not touch. Pre-existing.

```
$ bun run test:web
Test Files  89 passed (89)
     Tests  1298 passed (1298)
  Duration  52.46s        # exit 0
```

https://claude.ai/code/session_01Qa1QZe2qHEVaYJqweThoEY

### DIFF
--- a/apps/web/app/(public)/layout.tsx
+++ b/apps/web/app/(public)/layout.tsx
@@ -2,10 +2,9 @@ import type { ReactNode } from "react";
 import { AppShell } from "@/features/shell";
 
 /**
- * The public pages render in the same frame as the rest of the app. They used to
- * switch chrome on the session — a rail for members, a wordmark bar for everyone
- * else — but the designed rail carries its own signed-out state, so there is one
- * frame now and visitors keep the whole of it.
+ * The public pages render in the same frame as the rest of the app: the designed
+ * rail carries its own signed-out state, so there is one frame and a visitor
+ * keeps the whole of it rather than being given different chrome.
  */
 export default function PublicLayout({ children }: { children: ReactNode }) {
   return <AppShell>{children}</AppShell>;
--- a/apps/web/app/api/user/transcript/route.spec.ts
+++ b/apps/web/app/api/user/transcript/route.spec.ts
@@ -80,7 +80,7 @@ describe("a signed-out visitor", () => {
   /**
    * The change this route exists to make. The artboard has a guest read a
    * transcript and meet the account at the *keep* step
-   * (`docs/design_ref/2026-09-06/Course Community - Taken Courses.dc.html:1305`),
+   * (`docs/design_ref/2026-09-06/Course Community - Taken Courses.dc.html`),
    * which needs a parse that answers without a session — and which is safe only
    * because this route writes nothing.
    */
--- a/apps/web/app/api/user/transcript/route.ts
+++ b/apps/web/app/api/user/transcript/route.ts
@@ -71,7 +71,7 @@ const UNATTRIBUTED = "unattributed";
  * **A signed-out visitor may parse, and may not write.** The artboard has a
  * guest drop a transcript in, read what came out, and meet the account at the
  * *keep* step — "Sign in to keep this list"
- * (`docs/design_ref/2026-09-06/Course Community - Taken Courses.dc.html:1305-1308`)
+ * (`docs/design_ref/2026-09-06/Course Community - Taken Courses.dc.html`)
  * — and that flow is implementable only because parsing and writing are two
  * calls. This one is open; `transcript.confirm` stays a `protectedProcedure`
  * and is the only thing in the flow that touches a row. Nothing a signed-out
--- a/apps/web/app/globals.css
+++ b/apps/web/app/globals.css
@@ -137,25 +137,23 @@
   --cc-btn-fg: #ffffff;
   --cc-info: rgba(23, 81, 166, 0.07);
 
-  /* `--cc-warn-*` is the **review draft** family. It used to be documented here
-     as "the nudge to write a review", which #127 §4 flags as wrong: #88 showed
-     the actual nudge is `--cc-btn` on `--cc-surface`, and nothing amber appears
-     on it.
+  /* `--cc-warn-*` is the **review draft** family, not "the nudge to write a
+     review" — the nudge is `--cc-btn` on `--cc-surface` and nothing amber
+     appears on it.
 
      The tint half is the draft's header. The Design System artboard names
-     `--warn` "Review draft header" in both `TINTS` and `BANNERS`, and
-     `Course Community - Workspace Pane.dc.html` paints exactly it in its
-     `isReview` branch — `--warn` behind the draft's title block, `--warnInk` on
-     its "Review draft" eyebrow, and `--warnBorder` + `--warnSolid` as the
-     border and opaque backing of the progress row that sticks beneath it.
+     `--warn` "Review draft header" in both `TINTS` and `BANNERS`, and the
+     Workspace Pane artboard paints exactly it in its `isReview` branch —
+     `--warn` behind the draft's title block, `--warnInk` on its "Review draft"
+     eyebrow, and `--warnBorder` + `--warnSolid` as the border and opaque
+     backing of the progress row that sticks beneath it.
 
-     `--cc-warn-btn`/`--cc-warn-btn-fg` are a separate pair, and the one place
-     the old wording was not wrong: the Design System demos them under "Status
-     text & buttons" as a **"Write a review"** button. They are an accent, not a
-     banner tint — the filled progress segments and the submit button in
-     `review-draft-panel.tsx`, and the Review Card's accent for a review that
-     was not happy (`Review Card.dc.html:51`). Neither appears anywhere in the
-     Workspace Pane's `isReview` branch. */
+     `--cc-warn-btn`/`--cc-warn-btn-fg` are a separate pair: the Design System
+     demos them under "Status text & buttons" as a **"Write a review"** button.
+     They are an accent, not a banner tint — the filled progress segments and
+     the submit button in `review-draft-panel.tsx`, and the Review Card's accent
+     for a review that was not happy. Neither appears anywhere in the Workspace
+     Pane's `isReview` branch. */
   --cc-warn: #fdf7ea;
   --cc-warn-border: #f0e6d2;
   --cc-warn-ink: #7a5309;
@@ -174,12 +172,11 @@
      six tokens are the client half of that contract and the palette can be
      re-skinned here without a data migration.
 
-     Mirrored, like everything above: the revised `cc-theme.css` names all six
+     Mirrored, like everything above: `cc-theme.css` names all six
      (`--nodeAurora` … `--nodeViolet`), and the Design System artboard renders
      them under "Community graph nodes". They are *unassigned* rather than
      absent — #68 §3 settled that a node with no configured appearance is
-     `--cc-brand`, and these are the palette personalisation will draw from once
-     a tier writer exists. */
+     `--cc-brand`, and these are the palette personalisation draws from. */
   --cc-node-aurora: #2f9e8e;
   --cc-node-ember: #c96a4a;
   --cc-node-frost: #2f5fa8;
@@ -190,9 +187,8 @@
   /* The panel-tint families. A tint, its border and its ink are three separate
      tokens because a banner needs all three and none of them is derivable from
      `--cc-success` / `--cc-danger`: dark states them as alpha over the page,
-     light as flat mixes that are not a percentage of anything. Anything
-     deriving one with `color-mix` is reading the palette from before these
-     existed (#127 §1). */
+     light as flat mixes that are not a percentage of anything. Do not derive
+     one with `color-mix`. */
   --cc-success-tint: #e9f3ef;
   --cc-success-tint-border: #cfe4de;
   --cc-success-ink: #1c6b60;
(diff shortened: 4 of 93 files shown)


harrietkallberg:
Closes #191

harrietkallberg:
Pushed `4b34520`, addressing #201's three false claims.

All three verified independently before touching anything:

1. **236px is the rail's width.** `app-shell.tsx` renders the rail `w-[236px]`; the Explore artboard's rail is `width:236px`; and the artboard applies the value as `margin-right:{{ searchBarMargin }}` on a row centred in the content column. That column spans `236 → W`, so its centre is 118px right of the viewport centre and a 236px right margin cancels exactly that — the bar lands on the **viewport** centre, not over the results.
2. **`RESULTS_FLOOR = 396`**, so the `max-w-[560px]` bar box is *wider* than the results column at the floor. The old premise fails there.
3. The "nothing to correct" conclusion rested on both.

The bullet now reads as suggested, stated as it is true on `dev` today. It does not pre-empt Part 2.

### `saved.tsx`'s `savedTopMargin` — checked, and wrong in its own way

Not assumed to mirror the Explore one. `savedTopMargin` is the **identical** expression — `(openSignal || workspaceTabCount > 0) ? "236px" : "0"` — and is also applied as `margin-right`, despite the name. But the falsehood there was different: the note claimed the app's arrangement "does the same job **exactly** rather than approximately".

It does not. The strip sits inside the column the pane narrows, so it loses `PANE_DEFAULT + WORKSPACE_GAP` = **522px** by default, and a moving amount as the reader drags (down to `PANE_MIN + WORKSPACE_GAP` = 374px), against the artboard's flat 236px. Different amounts, for different reasons — which is fine, and is now what the comment says. Its `resultsMax` claim was re-checked and **holds**: the artboard computes it and never reads it in markup (0 occurrences of `{{ resultsMax }}`).

Also dropped a stray "just lost 504px" earlier in the same file, which omitted the 18px gap and would have contradicted the 522px figure below it.

### Category sweep: numeric design-intent claims

I swept every comment in `apps/web` making a claim about a specific number that can only be falsified by arithmetic against another file. Two were false (both fixed above), one was imprecise (the 504px). **Everything else checks out**, verified rather than assumed:

| Claim | Check | |
| --- | --- | --- |
| `card-geometry.ts`: Saved's artboard `summaryMax` 57px "taller than this ramp ever reaches" | ramp maxes at 2 lines × `SUMMARY_LINE_HEIGHT` 19 = 38px; 57 = 3 × 19 | ✅ |
| `collection-detail.tsx`: "36px on the reorder buttons", shifting the collapse "by a fifth of the ramp" | `size-[26px]` button column + `gap-2.5` (10px) = 36; 36 / `CARD_RAMP_SPAN` 170 = 0.21 | ✅ |
| `hero-network.tsx`: comet "`5.4 * NODE_RADIUS` = 21.6px … 7.6px into a padded rect … 17px short of the text" | segment table ends at 5.4; `NODE_RADIUS` 4, `FEATHER` 10, `SAFETY` 25 → 21.6, 21.6−14 = 7.6, 25−7.6 ≈ 17 | ✅ |
| `neighbourhood-view.ts`: "about 45 000 world units to itself", "about 55 across the artboard's 1440x900 hero" | `RADIUS_STEP` 120 → π·120² ≈ 45 239; 1440·900 / (0.72²·45000) ≈ 55.6 | ✅ |
| `neighbourhood-view.ts`: "radius of 7.5 inside a 12.5px ring" | `hero-network.tsx` draws `7.5` and `ctx.arc(x, y, 12.5, …)` | ✅ |
| `landing.tsx`: "the artboard shifts the whole landing layer up by 40px", code uses ten | `landingShift: explore && !soft ? "-40px" : "0px"`; code has `y: -10` | ✅ |
| `confirm-dialog.tsx` / `alert-dialog.tsx`: "asking for 440px silently gets 384px" | `max-w-sm` = 24rem = 384px; `AlertDialogContent` still carries `data-[size=default]:sm:max-w-sm` | ✅ |

### The category itself

#201 is right that neither #191's rubric nor my brief names this class, and it is the one my pass was structurally blind to. A comment asserting a design-intent reason for a number restates nothing, narrates nothing, and is not history — so it lands squarely in #191's *keep* list as "a deliberate deviation from the design authority, with its reason", and a reviewer working the rubric will protect it. The only test that finds it is arithmetic across two or three other files, which is exactly the kind of check #191's own thread says should be a **test that fails** rather than prose nobody re-derives. `theme-tokens.spec.tsx` is the existing precedent; `236`, `RESULTS_FLOOR`, `PANE_DEFAULT` and `CARD_RAMP_SPAN` are the numbers that would most repay the same treatment.

### Checks

```
$ bun run lint
Checked 427 files in 630ms. No fixes applied.
Found 8 warnings.        # exit 0 — all in .agents/skills/neon-drizzle/, untouched

$ bun run test:web
Test Files  89 passed (89)
     Tests  1298 passed (1298)  # exit 0
```

Comments-only re-verified on both changed files by the compiled-output check: byte-identical after `removeComments`.

Issue #203, opened 2026-09-06
Landing hero: bring back node drift, travelling signals and burst-on-click, and make tier 3 mean something


harrietkallberg:
The landing hero was specified as a living graph and shipped as a still picture. This issue brings back the motion the `2026-09-06` export defines, and in doing so makes the third personalization tier mean something.

Every decision below is settled — grilled through with the product owner on 2026-09-06 and recorded in **ADR 0006** and `CONTEXT.md`. **Do not re-open them.** Where this contradicts a comment in the code, the comment is what is wrong; see "Rationale that must be rewritten".

## Why

`docs/design_ref/2026-09-06/Course Community - Landing.dc.html` specifies drifting nodes, signals travelling along edges, and a click that fans one out along every edge a node has. `Course Community - My Page.dc.html:715` ties tier 3 to exactly that: *"Refer friends to unlock a signal that goes out from your node."*

None of it was carried through. #68 retired the artboard's synthetic field for the real graph — correctly — and the implementation dropped the motion as a side effect, which was never asked for. So today the landing page is inert, and tier 3 (`signalStyle`, "Signal on click") — the axis a member reaches only by reviewing their **entire transcript** — pays out a static ring most people cannot distinguish from a slightly different dot. `CONTEXT.md` logged this as an open gap, not a settled position.

## What already exists

Do not rebuild these:

- **The tier-3 picker works.** `PERSONALIZATION_AXES` (`server/graph/appearance.ts:118`) has `signalStyle` at tier 3; `node-profile.tsx:191` renders the buttons; `graph.setAppearance` writes them and re-derives the gate. Nothing about the picker changes.
- **The data is already on the client.** `WindowNode` carries `signalStyle` (`server/graph/service.ts:83`), `ScreenEdge` carries endpoints and a `clearance` (`neighbourhood-view.ts:203`). **No server change of any kind is in scope.**
- **The keep-out geometry is all there.** `clearAt`, `lineClearance`, `pushClear` in `hero-keepout.ts`.
- **The three styles have static geometry.** `FADE_RINGS`, `COMET_SEGMENTS`, `DASHED_RING` in `hero-network.tsx:92-111`. These are kept — they become the reduced-motion form.

### 1. Anchored drift

Port `Course Community - Landing.dc.html:1200-1227` and `makeNode` (:716).

- Each node keeps a **home** at its projected, `pushClear`-ed position, and wanders inside a **±5px box** around it, reflecting off the walls.
[...]

Issue #201, opened 2026-09-06
Workspace tabs follow the reader between Explore and Saved, and the two tab strips start 118px apart


harrietkallberg:
Workspace tabs currently follow the reader between Explore and Saved, and the tab strip starts 118px lower on Explore than on Saved. Both are settled below: every decision in this issue was made with the artboards and the code open, and the reasoning is recorded so an implementer does not have to re-derive it.

**All line references are against `4fda3d3` on `dev`.** Verify before trusting; this repo has been bitten by stale citations four times.

---

### The defect

`workspace-storage.ts` keeps the open list under a **single** key:

```ts
const WORKSPACE_KEY = "cc.workspace.open";
```

Both hosts — `features/search/components/explore.tsx:85` and `features/saved/components/saved.tsx:146` — call `useWorkspacePane()`, which reads and writes that one key. Navigating Explore → Saved unmounts one host and mounts the other, which rehydrates the *same* list. Tabs opened in Explore appear in Saved and vice versa.

### What the design says

The artboards settle this and they are unambiguous:

- `Course Community - Explore.dc.html` and `Course Community - Saved.dc.html` each declare their **own** `tabs: []` in local component state.
- `cc-store.js` — the store the two artboards genuinely share, and which does carry saved courses and collections across pages — has **no tab state at all**.

So in the design a tab belongs to the page it was opened on. Tabs opened in Explore live in Explore; tabs opened in Saved live in Saved.

Note for anyone reading the artboard source: `closeAll()` in both files closes **popup menus** (`compareMenu`, `takenMenu`, `overflowOpen`), not tabs. `onOpen: () => { this.closeAll(); this.openTab(...) }` in the Explore artboard is "dismiss any open menu, then open the tab" — it is *not* tab replacement. This is easy to misread.

### The decision: per-route key, persistence kept

The artboards persist nothing. We are **not** copying that, and the reason is a promise we make in copy.

`AuthReasonDialog` tells the reader *"Your draft is held as it is — text, ratings and the course all stay put"*, and the long comment at the top of `workspace-storage.ts` exists entirely to keep that promise across a sign-in redirect. Dropping persistence to match the artboard would break it.

Scoping and persistence are independent, so we take both:

- Key becomes `cc.workspace.open.<scope>`, where scope is `explore` or `saved`.
- Explore → Saved shows no tabs. Saved → Explore brings Explore's tabs back.
[...]

PR #204, opened 2026-09-06
docs(graph): settle the hero's travelling signals in the glossary and an ADR

harrietkallberg:
Documentation only — no code. This is the settled half of #203, so that the implementation branch has something to cite instead of re-deciding.

## Why

The `2026-09-06` export specifies a living hero: nodes drifting in a small box around their own position (`Course Community - Landing.dc.html:1200-1227`), signals travelling along edges, and a click fanning one out along every edge a node has (`burst`, :1140). `Course Community - My Page.dc.html:715` ties the third personalization tier to exactly that — *"Refer friends to unlock a signal that goes out from your node."*

None of it was carried through. #68 retired the artboard's synthetic field in favour of the real stored graph, which was right, and the motion was dropped as a side effect of that, which was never asked for. The result is an inert landing page and a tier 3 — the axis a member reaches only by reviewing their **entire transcript** — whose reward is a static ring most people cannot tell from a slightly different dot.

`CONTEXT.md` recorded this as an open gap rather than a settled position: *"Closed by whatever gives the hero a bounded loop a signal can ride on."* The product owner has now settled it, and this is that closure.

## What changes

**`CONTEXT.md`**

- **Signal** rewritten as a trail travelling *along a backbone edge*, not "along a node". The `_Today_` clause is deleted outright — the gap it recorded is closed by the decision, not documented by it.
- **Signal style** added as its own entry, so *"the node owns the style; the edge carries the signal"* is stated rather than inferred. The export names this axis three different ways (`NODE_SIGNAL_STYLES`, `TIER_AXES`, `DOT_SIGNALS`) and the confusion is worth foreclosing.
- **Burst** added. It is user-facing copy on My Page, not just a function name in the artboard.
- **Backbone edge** gains one clause pointing at the ADR, so the glossary's own "not a friendship" line does not read as contradicting the hero.

**`docs/adr/0006-hero-signals-dramatise-the-community.md`** — new.

## The three things in the ADR worth reading before approving

**The frame-loop guarantee gets weaker, deliberately.** The old design's promise was "the scene is painted once". The new one is a loop gated on four conditions — hero visible, not paused, not hidden, motion allowed. That is the price of the feature and the ADR says so rather than glossing it.

**Every node signals; the tier picks only the style.** Gating the signal itself on tier 3 would have made the feature invisible on a community where #68 recorded personalisation as having close to zero writers.

**It reserves vocabulary for a feature that does not exist yet.** `docs/landing_docs/personal-community-viewport.md` defers a friend feature whose mechanism is a "wormhole" that *moves the camera* without touching anyone's placement or backbone connections. Since this PR spends the language of travelling light on the backbone edge, the ADR records the constraint that falls out: **a friendship must never be drawn as an edge on this canvas, and must never emit a signal.** Without writing that down now, the natural instinct later is a brighter edge, and the distinction dies quietly.

The ADR also records what was rejected and why — click-burst only, the artboard's `INSIGHTS` payload, and its `Math.random()` sampling — so none of the three gets re-proposed as an improvement in six months.

## Follow-up, not in this PR

#203 carries the implementation: `features/landing/lib/hero-signals.ts`, the drift, the three producers, the four trail geometries, pointer events on the canvas, the loop gates, and rewriting the three places that currently argue for the opposite decision (`hero-network.tsx:33-47`, `hero-network.spec.tsx:7-19`, `neighbourhood-view.ts:45-51`). Those rewrites belong with the code that makes them true, not here.

## Note for whoever merges

ADR **0007** (workspace tabs, #201) is being written concurrently on another branch and also touches `CONTEXT.md` — a **School** entry, in a different section. The numbering does not collide. Whichever of the two merges second needs a trivial rebase.

Closes nothing on its own. Refs #203.

https://claude.ai/code/session_015NfoMyi4cztYt5USkxSXhs

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -159,6 +159,14 @@ draws **signals** travelling along it, which ADR 0006 settles as a deliberate
 dramatisation and not a claim about the data.
 _Avoid_: friendship, connection, follow, link, relationship
 
+**Graph window**:
+The bounded slice of the **community graph** a surface draws: a limited set of
+nodes and the **backbone edges** that run between them. A member's window is
+centred on their own node and is the one place "neighbourhood" means; a
+visitor's is centred on the community origin. It is derived per read and never
+stored, and an edge with one end outside it is not part of it.
+_Avoid_: viewport, camera, subgraph, cluster
+
 **Node profile**:
 A node's appearance, stored separately from graph topology.
 _Avoid_: avatar, skin, theme
@@ -178,9 +186,11 @@ the style; the edge carries the signal.
 _Avoid_: signal type, trail, effect
 
 **Burst**:
-The fan of signals a node sends along every one of its **backbone edges** at
-once, because somebody clicked it. Every node bursts; the tier decides only what
-its signals look like, never whether they go.
+The fan of signals a node sends at once, along every **backbone edge** it has
+within the drawn **graph window**, because somebody clicked it. A node near the
+window's rim is attached to nodes outside it; those edges are not part of the
+window and the burst does not reach them. Every node bursts; the tier decides
+only what its signals look like, never whether they go.
 _Avoid_: explosion, broadcast, emit, ripple
 
 **Personalization tier**:
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -154,7 +154,9 @@ _Avoid_: parent, host, neighbour
 **Backbone edge**:
 The stored attachment between a node and its anchor. Its direction records
 placement history — newer to older — and the UI may draw it undirected. It is
-not a friendship and carries no social meaning.
+not a friendship and carries no social meaning. The landing hero nonetheless
+draws **signals** travelling along it, which ADR 0006 settles as a deliberate
+dramatisation and not a claim about the data.
 _Avoid_: friendship, connection, follow, link, relationship
 
 **Node profile**:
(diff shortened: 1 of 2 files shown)


harrietkallberg:
Valid finding, fixed in `cae1baf` — and it was pointing at something larger than the Burst wording.

**The facts check out.** `anonymise` in `apps/web/server/graph/service.ts` keeps an edge only when both endpoint tokens resolve, and `projectGraphWindow` drops it again if either endpoint is missing from the projected set. Both reads cap at 150 nodes and every node holds 3–5 anchors, so a node near the rim certainly has attachments the payload does not carry. "Every one of its backbone edges" was promising adjacency that never arrives.

**Why it could not be stated precisely.** `CONTEXT.md` had no term at all for the bounded slice — no match for window, viewport or bounded anywhere in the file. Meanwhile the code calls it `GraphWindow` / `WindowNode` / `WindowEdge` / `getPublicWindow` / `projectGraphWindow`, and `docs/landing_docs/personal-community-viewport.md` calls the same concept a bounded viewport. Two names in the tree for a first-class concept the glossary never defined, which is exactly the condition ADR 0003 exists to remove.

So rather than qualifying Burst with rendering vocabulary, the fix names the concept:

- **Graph window** added, placed with the topology terms it is built from. It also settles "neighbourhood" as the member's own window rather than letting it drift into a second concept.
- **Burst** now states the bound in the domain's own words and says plainly what happens at the rim: a node near the edge of the window is attached to nodes outside it, those edges are not part of the window, and the burst does not reach them.

**The second suggested remedy is declined.** Specifying a bounded incident-edge expansion contract would be a server change, which #203 puts explicitly out of scope, and it would place a fetch-and-render contract inside a glossary. The first option — bound the promise — was the right one; it just needed a domain term to bind to rather than the word "rendered".

One note on the score's framing: this is a documentation-only PR with no code in it, so "not safe to merge" overstates the risk. The imprecision was real and worth fixing, but nothing here could ship a defect on its own — the implementation lands separately under #203, whose engine already operates entirely on the projected `GraphWindowView` and so was never going to reach for an edge the window does not hold.

PR #213, opened 2026-09-07
refactor(my-page,shell): the reader's own name as the page title, and the upload route retired

harrietkallberg:
## What

My Page opened on **286px of its own chrome** — 39% of a 730px viewport before the first stat card. Three blocks stacked above the tabs, and two of them named the account:

| block | height | |
|---|---|---|
| top bar | 66 | kept — it does real work on the landing page |
| `PageHeader` | 89 | "My Page" / "Private to you." |
| `Identity` | 70 | 52px avatar, name, "Change photo" |
| gap + tab strip | 61 | |

`PageHeader` gains an optional `leading` slot, and `/profile` is the only route that passes one. The reader's name becomes the title, their picture sits beside it, and `At KTH since <year>` joins the subtitle.

**Measured on a 730px viewport:** header row 89px where the two blocks were 159px; tab strip y=245 → **y=175**; first stat card y=286 → **y=216**. 39% of the viewport down to 30%.

`/saved`, `/search` and `/about` are unchanged to the pixel (`top:92 left:326 w:1120` on both branches). The row is a flex container whose title block carries `flex-1`, so a lone child lays out exactly as the block element it replaces, and `gap` only applies between items.

## The uploader is gone

`users.image` is a **Better Auth column**, and this app has Google and GitHub configured as social providers — both populate it on sign-in at no cost to us. The uploader existed only so a magic-link account could fill the same column, for a picture rendered in exactly one place that nobody but its owner ever sees (reviews carry a user id and no surface renders a reviewer's name).

Retired: a 77-line hardened route, its 146-line spec, `uploadProfilePicture`, the `@vercel/blob` dependency, `BLOB_READ_WRITE_TOKEN`, and `updateImage` through the user service and repository.

Accounts that already uploaded **keep** the image stored on their row — this removes the ability to set a new one, not the value. If the Blob store is ever deleted those URLs stop resolving, and Radix's `AvatarFallback` renders initials whenever an image fails to load, so they degrade rather than break.

## Initials moved to `lib/initials.ts`

Deleting `identity.tsx` would have left the rail's copy of the rule alone in the codebase. The rail already argued that a name and the initials beside it must fall through the same fields in the same order; both surfaces now compute them from one function.

## Two consequences, stated on purpose

- `PageHeader` is `hidden @3xl/shell:block`, so **the identity row is desktop-only now**. Mobile keeps the top bar's page title, which is what named the page there anyway (`pageTitleFor(pathname)`).
- This **merges two blocks `docs/design_ref/2026-09-06/Course Community - My Page.dc.html` draws separately**. The artboard governs, so the departure is recorded at the call site: it draws no avatar in that block at all, and the vertical budget it assumes is not the one a laptop has.

## Verification

- `bun run lint` — clean (the 8 warnings are pre-existing, all in `.agents/skills/`)
- `bun run typecheck` — exit 0
- `bun run test:web` — **93 files, 1390 tests, all passing**; 8 new tests cover the `leading` slot and `initialsOf`
- Rendered and measured in Chrome against `dev` on a second port, signed in, including confirming the real `lh3.googleusercontent.com` avatar renders

https://claude.ai/code/session_01UpU2R5YQWtbx5dREiaUETp


harrietkallberg:
Fixed in 3391e17 — thanks, the finding was correct and its file list was short by three.

**What you named:**
- `README.md` — `BLOB_READ_WRITE_TOKEN` removed from the required env vars
- `AGENTS.md` / `CLAUDE.md` — the multipart bullet now names `/api/user/transcript`, the only multipart route left, and states where `users.image` actually comes from (the Google/GitHub profile at sign-in; nothing in this app writes it)
- `docs/adr/0002` — annotated, see below

**Also stale, not in your list:**
- `.claude/agents/security.md` named `profile-picture/route.ts` as *the* upload surface to audit, and `BLOB_READ_WRITE_TOKEN` as a secret to guard — a security reviewer would have been sent to a file that no longer exists
- `.claude/agents/reviewer.md` held that route up as the pattern to copy for multipart
- `apps/web/server/http/body-cap.ts` explained its own existence with it

Both agent files now point at the transcript route.

**Two kept on purpose:**

ADR 0002's bullet is annotated rather than rewritten. The decision it records is "multipart bypasses tRPC", which still holds — only the example retired. That matches how the same ADR already handles its removed `docker-compose.yml`. ADR 0001's mention sits inside a description of NestJS body parsing, an architecture ADR 0002 replaced outright, so it reads as history.

**One correction found while writing the replacement:** my first draft of the security bullet said the transcript route "checks session". It does not — a signed-out visitor may parse a transcript and meets the account at the *keep* step. What bounds the open path is a per-caller and all-callers rate limit plus a parser-slot gate, checked before the body is read. The bullet says that now, since a reviewer told the route was session-gated would look straight past what actually holds it up.

Checks after the change: lint clean (8 pre-existing warnings, all in `.agents/skills/`), typecheck exit 0, 93 files / 1390 tests passing.

PR #223, opened 2026-09-07
docs: the placement backfill in the quick reference, and a shared graph window

harrietkallberg:
Follow-up to #222, which shipped `bun run backfill:placements` and made every graph window centred on the community origin — but left three documents describing the world before it.

## Quick reference

`CLAUDE.md` and `AGENTS.md` list one backfill. There are two now, and they are independent: one gives accounts a graph node, the other gives contributors the personalization tier they already earned. An agent that cannot see the second one will reach for SQL.

## Domain language

`CONTEXT.md` is where the words are settled, so a stale entry there is worse than a stale comment — it is the thing other documents are checked against.

**Graph window** said *"a member's window is centred on their own node and is the one place 'neighbourhood' means; a visitor's is centred on the community origin."* Both halves are now false. Every window is origin-centred, and a member's window is a visitor's window with one node flagged as theirs. "Neighbourhood" no longer names anything, because no window is centred on anybody — it moves to _Avoid_.

**Camera** is added as its own term, because the change split one idea into two. The window is *which nodes* a surface draws; the camera is *where it puts them*, and only the camera knows who is reading. Keeping them one word is what would let "just centre it on the viewer" back in.

**Community graph** gains one clause: not rebuilt per visit *and not rebuilt per reader*. That is the property the whole change exists to protect — two friends comparing screens see the same people in the same arrangement.

Docs only; no code, no tests affected.

https://claude.ai/code/session_01DxpL1ED7zRZLf8JvaC1jdS

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -13,6 +13,7 @@ Bun workspace monorepo (`workspaces: ["apps/*"]`). The only app is `apps/web`: N
 - DB: `bun run db:push` / `bun run db:generate`
 - Ingest: `bun run ingest` (optional `--test`)
 - Backfill earned personalization tiers: `bun run backfill:tiers` (idempotent; only ever raises)
+- Backfill missing graph nodes: `bun run backfill:placements` (idempotent; places app users who have none)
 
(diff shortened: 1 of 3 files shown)


harrietkallberg:
Superseded — applied directly on dev instead (73fb2d8).

PR #238, opened 2026-09-08
fix(transcript): accept degree-project course codes

### LINKED ISSUE #230: Transcript parser drops KTH degree-project codes ending in X
## What happens

The Ladok transcript parser does not recognise KTH degree-project codes such as `DA232X` and `SF279X`.

An X-only transcript reaches "No completed courses were found in this transcript." A mixed transcript is worse: its ordinary courses are returned while the degree project is silently omitted, so the reader can confirm an incomplete taken-course list without being told anything was lost.

## Why

All three course-code patterns in `apps/web/server/ingest/transcript/parse.ts` require four digits before an optional letter:

```ts
/^[A-ZÅÄÖ]{2,3}\d{4}[A-Z]?$/
```

KTH degree-project codes use three digits followed by `X`. The checked-in catalogue corpus at `apps/web/server/derive/__fixtures__/corpus.json` contains **35 real examples**, including `AB100X`, `DA232X`, and `SF279X`.

Because `startsRow` rejects the line, neither the full-row parser nor `ROW_HEAD` ever gets a chance to preserve it.

## Fix

Define one course-code source/pattern that accepts both shapes the catalogue contains:

- ordinary codes: two or three letters + four digits + optional suffix letter
- degree-project codes: two or three letters + three digits + `X`

Use that same definition for `COURSE_CODE`, `ROW`, and `ROW_HEAD`; do not widen the degree-project branch to arbitrary three-digit suffixes without catalogue evidence.

Update the nearby "four digits" documentation so it describes both real shapes.

## Tests

Add parser coverage for:

- a complete `DA232X` row with credits, grade, and date
- a wrapped/fallback `SF279X` row
- a mixed transcript where the X course is retained alongside an ordinary course
- an X-only transcript, proving it no longer throws the no-courses error
- an unrelated malformed token that must still not start a row

## Done when

- All 35 catalogue `…X` code shapes are accepted by the parser.
- Ordinary transcript parsing is unchanged.
- Degree-project rows cannot be silently omitted.
- `bun run lint`, `bun run typecheck`, and `bun run test:web` pass.

FilipDimitrijevic97:
## Summary

- share one transcript course-code pattern across row detection and parsing
- accept KTH degree-project codes with three digits followed by `X`
- retain the existing ordinary four-digit course-code behavior without widening other three-digit suffixes

## Validation

- [x] transcript parser tests — 18 passed
- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] independent code review — no findings
- [ ] `bun run test:web` — 1,460 passed; three unrelated PDF extraction tests could not resolve `unpdf` in the isolated local dependency install

No database migration or schema change.

Closes #230

### DIFF
--- a/apps/web/server/ingest/transcript/parse.spec.ts
+++ b/apps/web/server/ingest/transcript/parse.spec.ts
@@ -246,6 +246,85 @@ describe("parseLadokTranscript", () => {
     ]);
   });
 
+  it("reads a complete degree-project row", () => {
+    const transcript = [
+      "Code Name Scope Grade Date Note",
+      "DA232X Degree Project in Computer Science 30.0 hp P 2026-06-05 1",
+      "Summation",
+    ].join("\n");
+
+    expect(parseLadokTranscript(transcript)).toEqual([
+      {
+        courseCode: "DA232X",
+        courseName: "Degree Project in Computer Science",
+        credits: 30,
+        grade: "P",
+        completedOn: "2026-06-05",
+      },
+    ]);
+  });
+
+  it("reports a wrapped degree-project row through the fallback parser", () => {
+    const transcript = [
+      "Code Name Scope Grade Date Note",
+      "SF279X Degree Project in Applied and",
+      "Computational Mathematics",
+      "Summation",
+    ].join("\n");
+
+    expect(parseLadokTranscript(transcript)).toEqual([
+      {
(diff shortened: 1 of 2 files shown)


harrietkallberg (approved):

harrietkallberg:
lgtm!

PR #239, opened 2026-09-08
fix(taken): validate manual attendance year

### LINKED ISSUE #231: Manual course entry accepts attendance years the API rejects
## What happens

The **Add a course by hand** dialog accepts any four digits in its optional Year field. Values such as `1899`, `2201`, and `0000` enable submission and are sent to the server, where `taken.add` rejects them because its schema allows only 1900–2200.

The reader receives the parent's generic save failure instead of an inline explanation, even though the same feature's inline row editor already knows and applies the correct range.

## Why

`apps/web/features/taken/components/add-taken-course-dialog.tsx` parses the value with:

```ts
const parsedYear = /^\d{4}$/.test(year) ? Number(year) : null;
```

The input strips non-digits and has `maxLength={4}`, but neither the field nor `submit` checks the server's numeric range.

`apps/web/features/taken/lib/taken-rows.ts` already exposes `parseYear`, which returns:

- `null` for a blank optional value
- a year for 1900–2200
- `undefined` for malformed or out-of-range input

## Fix

Use the existing `parseYear` contract in the manual-add dialog rather than maintaining a second, weaker parser.

- Blank remains valid and sends `null`.
- A valid 1900–2200 year submits normally.
- Invalid or out-of-range input stays in the dialog, does not call `onAdd`, and gets a clear inline error.
- Preserve the reader's draft so they can correct it.

## Tests

Cover blank, 1900, 2200, 1899, 2201, and a partial year. Assert invalid values never call `onAdd` and valid boundaries send the parsed number.

## Coordination

PR #228 recently changed this dialog/search flow and its parent spec. Work from current `dev` and retain its shared search-field behavior.

## Done when

- The dialog and `taken.add` accept the same year domain.
- Out-of-range input is refused before the mutation with actionable copy.
- Blank remains a valid optional year.
- `bun run lint`, `bun run typecheck`, and `bun run test:web` pass.

FilipDimitrijevic97:
## Summary

- reuse the taken-course year parser in the manual-add dialog
- accept blank years and the 1900/2200 boundaries while refusing partial or out-of-range values before mutation
- preserve invalid drafts and associate clear inline validation copy with the field

## Validation

- [x] Taken UI tests — 105 passed
- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] independent code review — no findings
- [ ] `bun run test:web` — 1,460 passed; three unrelated PDF extraction tests could not resolve `unpdf` in the isolated local dependency install

No database migration, schema change, repository change, or deletion.

Closes #231

### DIFF
--- a/apps/web/features/taken/components/add-taken-course-dialog.tsx
+++ b/apps/web/features/taken/components/add-taken-course-dialog.tsx
@@ -12,7 +12,7 @@ import {
   useDebouncedQuery,
   useSearchCourses,
 } from "@/features/search";
-import { creditsLabel, type TakenEdits } from "../lib/taken-rows";
+import { creditsLabel, parseYear, type TakenEdits } from "../lib/taken-rows";
 
 /** The credits a KTH course usually carries, as the artboard's own chips. */
 const CREDIT_CHOICES = [6, 7.5, 9, 15];
@@ -68,6 +68,8 @@ export function AddTakenCourseDialog({
   const search = useSearchCourses({ q: debounced });
   const results = search.data?.results ?? [];
   const alreadyTaken = new Set(takenCourseCodes);
+  const attendanceYear = parseYear(year);
+  const yearInvalid = attendanceYear === undefined;
 
   function reset() {
     setQuery("");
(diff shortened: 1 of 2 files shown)


harrietkallberg:
lgtm!

harrietkallberg (approved):

